Issue #11, opened 2014-11-12
Non-actionable warning in console: You tried to return focus to null but it is not in the DOM anymore


coderberry:
I am seeing a lot of messages in the console: `You tried to return focus to null but it is not in the DOM anymore`. I am unsure how to make this warning go away.

misuba:
I notice the stuff in that commit doesn't seem to be in the code anymore... and I'm having this issue.

glifchits:
@misuba I'm with you there. I get the warning almost every time state most places in my app change, even when the modal is not open.

jeffutter:
I am seeing this too. @mzabriskie can we reopen this issue?

jccode:
Still seeing this warning too.  :(

Reproduce steps:

1. Construct 2 modals, named it modal_A and modal_B. 
2. Open modal_A (with one button). 
3. In modal_A 's click handler, try to close modal_A and open modal_B.
4. When you try to close modal_B, you can see the warning messages in console window.

mrkvngrkn:
I'm getting this warning since I've upgraded to React v16

diasbruno:
@jccode @mrkvngrkn Please follow issue #577. We are discussing how we are going to handle this properly.

PR #16, opened 2015-01-19
[fixed] ModalPortal's componentWillReceiveProps

arasmussen:
For some reason componentWillReceiveProps is being called with identical props. We need to compare previous props with new props to make sure state is changing from closed to open before calling this.open(). Otherwise we're calling this.open() even when state isn't changing. Same with this.close().


jeffutter:
It seems like this fixes #11 for me and does not seem to break anything. Could we get this merged?

ryanflorence:
v0.1.0 published with this fix

jeffutter:
@ryanflorence Awesome! Thank you

PR #22, opened 2015-03-18
Shim the possibly-absent HTMLElement

misuba:


Josh-a-e:
+1 This resolves the issues we we're having around rendering this server side and for testing in a node environment.

petebrowne:
Also, IE8 does not have an HTMLElement

PR #28, opened 2015-04-10
Use classnames instead of react/lib/cx

peterjmag:
classSet / cx will be removed in a future version of React. 

This also removes the following warning when running the test suite:

    WARN: 'Warning: React.addons.classSet will be deprecated in a future version. See http://fb.me/react-addons-classset'

See also:
https://github.com/facebook/react/commit/b6980ab9809fc3ddc68c20129618f0414d49b2a4
https://www.npmjs.com/package/classnames


fdubost:
:+1:

PR #31, opened 2015-04-30
Check if modalElement exists in handleFocus.

maisano:
A race condition happens if you open a new window before closing the modal, as opening a new window has the side effect of blurring the current window.


jacobangel:
+:100:

PR #36, opened 2015-05-08
Allow component to render without DOM libs

joeybaker:
This is just a quick hack that enables the component to not throw when being rendered on the server


mzabriskie:
Duplicate of #22

joeybaker:
nifty. thanks!

PR #38, opened 2015-05-11
[fixed] Allow version 2.x of classnames dependency

danez:
The api has not changed only some rare edge cases which are not used here.

Without the change this library can't be used with classnames 2.x

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   },
   "peerDependencies": {
     "react": ">=0.12.0",
-    "classnames": "^1.2.0"
+    "classnames": "1.2 – 2"
   },
   "tags": [
     "react",


ncuillery:
Why not `>=1.2.0` (to be quiet if classnames 3.x is released ?)

Anyway: :+1: I need this

danez:
Yeah, would also work, its up to one of the maintainers to decide if the want to allow any newer version.

astephenb:
:+1:

lge88:
As far as I know, peer dependencies are for plugins-host related modules. Could anyone explain why making `classnames` a peerDependency instead of dependency?

remimarenco:
:+1: , I need this too ;)

Spy-Seth:
I need it !


:+1:

prettynatty:
:+1:

ascrazy:
:+1:

Galathius:
:+1:

mzabriskie:
classname module has been updated.

PR #41, opened 2015-06-01
Add class to body when modal is open fixes #29

### LINKED ISSUE #29: [Feature Request] Add class to app element when modal open
We would like to lock the app from scrolling when the modal is open. Currently we have a wrapper around `Modal` to set the class on our app element based on the `isOpen` prop.

KyleAMathews:
This lets you freeze the background on mobile.


astephenb:
Thanks @KyleAMathews. I was thinking that the class would be added to the app element that was set by `Modal.setAppElement(element)`

When looking that how to implement that, it wasn't immediately clear what would be needed since the `ariaAppHider` module is the only unit that cares about the app element (from what I can tell). My initial thought was to expose the app element via the `ariaAppHider`.

KyleAMathews:
That was my initial thought—to add it to the app element—but there's no reason the class should be put there and it's more generally useful if put on the body class.

astephenb:
That makes sense. I can add some test coverage of this.

KyleAMathews:
That'd be great!

mzabriskie:
Some of your changes were already merged with a different PR, so I hand picked your body class changes. Thanks for the PR.

PR #42, opened 2015-06-07
Moves classnames to dependencies

claydiffrient:
This makes it a bit more compatible with the way peerDependencies will work in `npm@3` later down the line.
It also makes it so that consumers of the lib can have another version of classnames at the app level while
not requiring this particular lib to require a version bump.  Plus, this lib sorta depends on it and should
it need to be locked down to a particular version, this lib should manage that rather than expect it to
happen at the app level.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,9 +39,11 @@
     "uglify-js": "2.4.15",
     "webpack-dev-server": "1.6.5"
   },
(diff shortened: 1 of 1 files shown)


ivan-kleshnin:
+1

PR #54, opened 2015-07-26
Move injected JS styles to real CSS file

alekstorm:
Injecting a fake `<style>` tag made it difficult to override the styles, and made them impossible-to-reach for static analysis tools.


daisuke-3:
awesome.
Looks good this way.
I also want to override the style easily.
:+1:

mzabriskie:
After removing all the `injectCSS` calls, how are the examples getting their styles?

claydiffrient:
I think with the changes to inline styles that occurred in v0.5.0 this seems unnecessary now.  Thanks for the PR, but I'm going to close it for now.

PR #55, opened 2015-07-28
fixed issue displaying modal quickly after closing it

Josh-a-e:
This solves an issue I encountered where if I am 
- using the `closeTimeoutMS` prop, 
- causing the modal to close
- opening the modal again before the specified number of milliseconds has passed

This results in the `--before-open` classes not getting removed, which causes the modal to be present but invisible in my case (due to its opacity being set to zero by the `--before-open` classes).

resulting html of modal opened immediately after being closed

I imagine this is a not entirely uncommon: I had presented a modal to confirm an action and then on confirmation (after a http request had completed) presented another modal to notify the user of outcome (success / failure). The issue being the http request took less than the 400ms I had set on `closeTimeoutMS`.


mzabriskie:
@Josh-a-e this looks like a good fix. Would you mind resolving the conflicts, and we'll get it merged?

Josh-a-e:
@mzabriskie I've merged in the latest master - thanks!

claydiffrient:
@Josh-a-e Could you squash the 6 commits down to 1 (to keep our history a bit cleaner, and make it easier to reason about) and add a spec?  Then I'm happy to merge.

diasbruno:
This issue seems to be fixed on newer versions. If it still happens, please reopen this PR.

PR #70, opened 2015-09-23
added suport for isomorphic rendering

dinodsaurus:
I just added check if the component is running in browser or node.
Before when rendering on server there was an error 
```
/Users/dino/Documents/dino/react-modal/lib/helpers/ariaAppHider.js:4
if (document) {
    ^

ReferenceError: document is not defined
```

### DIFF
--- a/lib/helpers/ariaAppHider.js
+++ b/lib/helpers/ariaAppHider.js
@@ -1,7 +1,4 @@
-var _element = "";
-if(typeof window !== 'undefined'){
-  _element = document.body;
-}
+var _element = typeof document !== 'undefined' ? document.body : null;
 
 function setElement(element) {
   if (typeof element === 'string') {
--- a/lib/helpers/ariaAppHider.js
+++ b/lib/helpers/ariaAppHider.js
@@ -1,4 +1,7 @@
-var _element = document.body;
+var _element = "";
+if(typeof window !== 'undefined'){
+  _element = document.body;
+}
 
 function setElement(element) {
   if (typeof element === 'string') {
@@ -39,4 +42,3 @@ exports.setElement = setElement;
 exports.show = show;
 exports.hide = hide;
 exports.resetForTesting = resetForTesting;
-


tkirda:
Yes, that is needed! Now it fails when rendering isomorphic app. I suggest changing first line into:

```javascript
var _element = typeof document !== 'undefined' ? document.body : null;
```

There is no point setting _element to string.

dinodsaurus:
Thanks for this, I updated my pull request :+1:

Josh-a-e:
Would love to see this merged in :smile:

sebslomski:
Please merge this, it works perfectly.

ghislaindj:
Totally agree, this PR works great !

voronianski:
rackt guys @gaearon @mzabriskie pleas merge this line and bump npm version pleeeeeaseee :watch:

mzabriskie:
Has anyone tested this diff with a screen reader? The fix would work for server rendering. Just need to make sure that once we're on the client `_element` still gets `aria-hidden="true"` when modal is displayed.

tkirda:
It does not affect the behavior in any way, it's just to prevent exception when including file in nodejs runtime. Since `document` is not available. On a client it will behave the same way.

mzabriskie:
Sounds good. Thanks!

tkirda:
Sweet! Looking forward to see it on NPM.

voronianski:
@mzabriskie :+1:

Spy-Seth:
Any idea on when a new release will be?

mzabriskie:
I will update react to 0.14, and release today.

sahat:
Has this been published yet? I am still having this issue in version 0.5.

jhabdas:
+1

vonwao:
Still doesn't work with meteor.

ericntd:
It looks like the fix is included in 0.6.0.
But then 0.6.0 requires React 0.14 and I'm using React 0.13.3.
What else can I do besides crying and changing the source code of your library under node_modules folder?

dinodsaurus:
In my `package.json`I am loading it as this:
`"react-modal": "git+https://github.com/rackt/react-modal.git#920d421ed14b94b47bb32310eb691f763d67d464"`

PR #80, opened 2015-10-15
Remove lodash from browserify

ndreckshage:
**Do not merge!** Untested etc. Just for discussion.

Setting lodash.assign as a global in browserify config reduces the unminified build in dist from 49.5k to 16.1k. And the minified version from 20.6k to 11.3k. Non gzipped (...Github should show gzip size).

I almost didn't use the library + wrote my own because the numbers looked crazy :smile:. Maybe just add size to readme? This is tiny gzipped, minified, and without lodash (4k). Thanks!


claydiffrient:
@ndreckshage Thanks for opening this up for discussion.  I'm thinking we can definitely work out a change that might reduce the size of the bundle.  I'm hoping that perhaps we might be able to move completely off browserify and onto webpack completely with some of its handling of bundles.  I think that if we just required the lodash.assign function, we could likely get rid of most of the lodash bulk.

claydiffrient:
@ndreckshage Would you mind looking at #144 and letting me know your thoughts?

diasbruno:
Just pinging. 
@ndreckshage @claydiffrient browserify was removed in #159. Any thoughts?

claydiffrient:
I'm going to close this based on the changes that were made to enhance the build process and reduce the minified build size in #159

PR #100, opened 2015-11-06
[fixed] Custom classnames override default styles

claydiffrient:
This makes it so if you provide a custom className it will override
all the default styling, relying solely on the style props you pass
in and the className itself.

Prior to this commit, providing a className or overlayClassName
doesn't really have any effect because the style attribute has higher
precedent when applying styles.

### DIFF
--- a/lib/components/ModalPortal.js
+++ b/lib/components/ModalPortal.js
@@ -175,16 +175,19 @@ var ModalPortal = module.exports = React.createClass({
   },
 
   render: function() {
+    var contentStyles = (this.props.className) ? {} : defaultStyles.content;
+    var overlayStyles = (this.props.overlayClassName) ? {} : defaultStyles.overlay;
+
     return this.shouldBeClosed() ? div() : (
       div({
         ref: "overlay",
         className: this.buildClassName('overlay', this.props.overlayClassName),
-        style: Assign({}, defaultStyles.overlay, this.props.style.overlay || {}),
+        style: Assign({}, overlayStyles, this.props.style.overlay || {}),
         onClick: this.handleOverlayClick
       },
         div({
           ref: "content",
-          style: Assign({}, defaultStyles.content, this.props.style.content || {}),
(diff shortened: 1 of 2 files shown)


jackofseattle:
:+1: 
Good idea. I was trying to think of a way to do this when I first added in the inline styles. I think this is a very good option.

claydiffrient:
@mzabriskie Your thoughts?

sergeyk:
:+1:

claydiffrient:
I've been using this for a bit now... the only caveat that I've come across is that no default styles get applied... so you need to pass in the the default styles that may/may not matter to you into the styles prop when creating the modal.

chielkunkels:
Wouldn't it be better to just have a prop, say `disableStyles`, to disable the default styling, rather than depending on something that's arguably a little bit "magic"? That way, you can still specify custom classnames as well retain default styling.

claydiffrient:
@chielkunkels So this approach doesn't attempt to disable default styles (or at least that's not the end goal).  It is to make it so that class names have some value again.  The current implementation on master completely relies on the inline styles that are passed in.  Because of style precedence however, styles from class names are never applied.

Disabling the default styles just happens to be a byproduct of this approach.  It's not the intention of the PR though.

joefiorini:
:+1: @mzabriskie is this good to merge?

omarish:
:+1: @mzabriskie this would be huge.

edbury:
:+1: @mzabriskie would be so pumped on this.

idmontie:
What is the status of this PR? Is there anything holding it up?

ChrisZieba:
Is this project dead, why aren't solid PR's getting merged in?

aarosil:
+1, thank you! Can we get npm version?

claydiffrient:
@aarosil I'm hoping to get an npm version pushed out by the end of the week.  I'm hoping to get as many PRs merged in and issues closed before cutting the release.  So I'm hoping that the end of the week will give me enough time to take care of most of the current backlog.

PR #101, opened 2015-11-06
Bugfix/two or more modal open lose class on body

### LINKED ISSUE #99: CSS class .ReactModal__Body--open isn't added to body
The CSS class .ReactModal__Body--open isn't added to body, so I can't add overflow: hidden to the body in order to prevent scrolling behind the modal.

I'm using version 0.5.0.

roth1002:
Resolve #99  and  Resolve className of modal's refs test error on chrome

```
Chrome 46.0.2490 (Mac OS X 10.11.1) Modal supports custom className FAILED
    TypeError: modal.portal.refs.content.className.contains is not a function
        at Context.<anonymous> (/var/folders/zy/rzg9tmp57c5gl76s51rwy1c40000gn/T/6973524eb4f1221eeef8f0d767335739.browserify:22117:47 <- specs/Modal.spec.js:98:0)
Chrome 46.0.2490 (Mac OS X 10.11.1) Modal supports overlayClassName FAILED
    TypeError: modal.portal.refs.overlay.className.contains is not a function
        at Context.<anonymous> (/var/folders/zy/rzg9tmp57c5gl76s51rwy1c40000gn/T/6973524eb4f1221eeef8f0d767335739.browserify:22123:47 <- specs/Modal.spec.js:104:0)
```


jrf0110:
Anybody know the status of this? I'm really needing multiple modals :/

evoyy:
From what I can tell, this PR doesn't actually fix the underlying problem, it only allows for a workaround by allowing you to specify a custom class name to be applied to the body instead of ReactModal__Body--open. This will not scale with many modals on the page and will lead to a maintenance mess.

deadbeef404:
@evoyy A solution to that problem would be to have each modal have it's body-class start with "ReactModal__Body--open". Once that's done, the following CSS rule will allow you to toggle body styles when at least one modal is open. For example, you could disable the body scrolling. 

body[class^="ReactModal__Body--open"] {
  overflow: hidden;
}

You could add as many modals as you want to the body tag, and as long as at least one exists, the selector should adequately do the job.

Hi, @mzabriskie. Seeing as you seem to do most releases for this project, would you have any idea when this could get merged? Thank you!

claydiffrient:
@roth1002 @deadbeef404 I think this needs a bit more discussion and thought before we go down this route.  I'd like to think about it a bit.  As is, this PR cannot be merged anyway because it modifies things within the dist/ directory.  Those are generated with our release script and should not be modified in PRs.

dhruvparmar372:
I've been facing the same issue where the body className doesn't stick if 2 or more modals are being rendered in parellel. This is workaround I'm using for now
```
renderPortal: function(props) {
  if (props.isOpen) {
    elementClass(document.body).add('ReactModal__Body--open');
    setTimeout(function () {
      elementClass(document.body).add('ReactModal__Body--open');
    })
  } else {
    elementClass(document.body).remove('ReactModal__Body--open');
  }
  //other stuff
}
```
This ensures that even if 1 modal is open the `ReactModal__Body--open` is attached to the body element. Thoughts/Suggestions are welcome

claydiffrient:
I don't quite understand the use case for this.  Why would you have two modals open at once?  It seems that that would provide a terrible user experience.

I can't figure out the accessibility story with it either, because one modal will have focus trapped within it, rendering the second modal inaccessible completely.

dhruvparmar372:
As terrible as it sounds it does have a lot of valid use cases. Just to point one out 
Open any post with a photograph on facebook by clicking on the photograph. The post opens up in a modal with a close button. Now click on the list of likers. Another modal opens on top of the previous one. 

I myself had the same opinion about modal on a modal until stumbling on such use cases which actually don't look terrible in any way and don't affect accessibility too. 

User experience discussions apart as a library/plugin I don't think terrible use case is a valid point for avoiding a functionality piece which in theory is possible. After a little bit of digging around I stumbled upon this https://github.com/qimingweng/react-modal-dialog which handles this piece quite elegantly.

claydiffrient:
@dhruvparmar372 Thanks!  I don't mean to avoid the functionality.  I'm happy to make sure that it works and we focus primarily on the accessibility story since that is the primary function of the library.  Having a modal that opens a "sub-modal" is a completely valid use case.

I'm going to think through the implementation here and I think we can totally come up with something that works well.

briankadar:
@claydiffrient: I ran into this problem recently as well.

The issue presents itself when there is only one Modal open (e.g. `isOpen="true"`), but there are two or more <Modal/> components in the render tree (even if only one is visible). On `componentWillReceiveProps`, `this.renderPortal` is blindly called on all of the instances, causing the non-open modal to remove the body class.

I worked around this by:
* Composing <ReactModal> into my custom modal component
* Using a new, custom class name
* Creating a simple system for multiple instances of modals to register their open state

Here's some adapted pseudo-code:

```js

// Basic system for keeping track of open modals.

let openModalCount = 0;
let flushDelay;

const enqueueFlush = () => {
  if (flushDelay) {
    clearTimeout(flushDelay);
  }

  flushDelay = setTimeout(flushModalClass, 0);
}

const registerModalOpen = (isOpen) => {
  isOpen ? openModalCount++ : openModalCount--;
  enqueueFlush();
}

const flushModalClass = () => {
  flushDelay = null;
  if (openModalCount === 0) {
    elementClass(document.body).remove('my-custom-modal-open-class');
  } else {
    elementClass(document.body).add('my-custom-modal-open-class');
  }
};

export default class CustomModal extends Component {

  static propTypes = {
    ...ReactModal.propTypes
  };


  componentDidMount() {
    if (this.props.isOpen) {
      registerModalOpen(true);
    }
  }

  componentWillReceiveProps(nextProps) {
    if (nextProps.isOpen !== this.props.isOpen) {
      registerModalOpen(nextProps.isOpen);
    }
  }

  componentWillUnmount() {
    if (this.props.isOpen) {
      registerModalOpen(false);
    }
  }

  render() {
    const { children, ...rest } = this.props;

    return (
      <ReactModal { ...rest }>
      	{children}
      </ReactModal>
     );
  }
}
```

package.json
```diff
@@ -13,10 +13,11 @@
     "example": "examples"
   },
   "scripts": {
-    "test:firefox": "scripts/test --browsers Firefox --single-run",
-    "test:chrome": "scripts/test --browsers Chrome --single-run",
```
diasbruno: +1 we should move this to a separate PR, so we can run something like `npm run test:firefox -- --single-run`. So, maybe, we can include `travis`.(?)

README.md
```diff
@@ -176,7 +177,12 @@ var App = React.createClass({
           onAfterOpen={this.afterOpenModal}
           onRequestClose={this.closeModal}
           style={customStyles}
+<<<<<<< HEAD
```
claydiffrient: Merge artifacts should be removed.

diasbruno:
Since `react-modal` can handle multiple modals since `refCount` was implemented (e579a0d - available on `v1.7.4`), I'll close this PR for clean up.

PR #105, opened 2015-11-22
Ability to anchor modals relatively to its context rather than just the body.

addnab:
This can be done just by setting the prop `relative = true`.

Here are the possible use cases:
- When you want a modal to appear on a particular part of the page.
- Multiple modals to appear on different parts of the page.

I don't think the current release solves these use cases already. Please correct me if I'm wrong.

### DIFF
--- a/lib/components/Modal.js
+++ b/lib/components/Modal.js
@@ -21,6 +21,7 @@ var Modal = module.exports = React.createClass({
   },
 
   propTypes: {
+    relative: React.PropTypes.bool,
     isOpen: React.PropTypes.bool.isRequired,
     style: React.PropTypes.shape({
       content: React.PropTypes.object,
@@ -34,6 +35,7 @@ var Modal = module.exports = React.createClass({
 
   getDefaultProps: function () {
     return {
+      relative: false,
       isOpen: false,
       ariaHideApp: true,
       closeTimeoutMS: 0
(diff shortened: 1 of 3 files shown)


evoyy:
I believe this should already be possible. If you look in the README is says:

````javascript
/*
By default the modal is anchored to document.body. All of the following overrides are available.

* element
Modal.setAppElement(appElement);

* query selector - uses the first element found if you pass in a class.
Modal.setAppElement('#your-app-element');
*/
````

However, I noticed that this is not working correctly. I have opened this issue https://github.com/rackt/react-modal/issues/107

addnab:
@evoyy I did notice the note on the README. But I found all it did was set/unset the `aria-hidden` attribute on whatever AppElement you provided it with.

evoyy:
I noticed that also. I think we will have to wait for clarification from the maintainers. I have also submitted a PR https://github.com/rackt/react-modal/pull/108 which makes the modal do what I had expected it to do.

addnab:
I don't find the approach of having to set the element you want to anchor it to as a prop very elegant. It would feel nicer to just plug the Modal Component into whichever element you'd like a modal to be over.

evoyy:
I agree. I removed that suggestion from my PR. I will be happy to go with your solution.

claydiffrient:
I'm not sure that this particular request fits in line with what this component is trying to accomplish.  This component is intended to be a very accessible modal.  I'm having a hard time figuring out how the use cases you mentioned work from a UX perspective.  

> Multiple modals to appear on different parts of the page.

Generally when a modal is open, the entire experience focuses on that modal being open, that's why everything behind it fades out.  That essentially means that only one modal should ever be open at a time.

> When you want a modal to appear on a particular part of the page.

As I mentioned earlier, a modal takes focus of the entire user experience and should essentially take up the entire page, rather than a particular section of it.

Thanks for the PR, but I'm going to close it because I don't think its that good of an idea at this point.  If there is significant interest in the future, perhaps we will reconsider.

addnab:
I understand. Thanks.

PR #108, opened 2015-11-26
[fixed] Restore Modal.setAppElement() functionality

evoyy:
Maybe I have misunderstood, but this commit will anchor the modal to the specified appElement, as I had expected it to from the README. It fixes https://github.com/rackt/react-modal/issues/107.

### DIFF
--- a/lib/components/Modal.js
+++ b/lib/components/Modal.js
@@ -7,12 +7,15 @@ var elementClass = require('element-class');
 var renderSubtreeIntoContainer = require("react-dom").unstable_renderSubtreeIntoContainer;
 
 var SafeHTMLElement = ExecutionEnvironment.canUseDOM ? window.HTMLElement : {};
+var AppElement = ExecutionEnvironment.canUseDOM ? document.body : {appendChild: function() {}};
 
 var Modal = module.exports = React.createClass({
 
   displayName: 'Modal',
   statics: {
-    setAppElement: ariaAppHider.setElement,
+    setAppElement: function(element) {
+        AppElement = ariaAppHider.setElement(element);
+    },
     injectCSS: function() {
       "production" !== process.env.NODE_ENV
         && console.warn('React-Modal: injectCSS has been deprecated ' +
@@ -43,7 +46,7 @@ var Modal = module.exports = React.createClass({
   componentDidMount: function() {
     this.node = document.createElement('div');
     this.node.className = 'ReactModalPortal';
-    document.body.appendChild(this.node);
+    AppElement.appendChild(this.node);
     this.renderPortal(this.props);
   },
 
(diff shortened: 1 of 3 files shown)


lib/components/Modal.js
```diff
@@ -7,12 +7,15 @@ var elementClass = require('element-class');
 var renderSubtreeIntoContainer = require("react-dom").unstable_renderSubtreeIntoContainer;
 
 var SafeHTMLElement = ExecutionEnvironment.canUseDOM ? window.HTMLElement : {};
+var AppElement = document.body;
```
azaharakis: This would throw an error on Isomorphic/Universal applications :(

evoyy: @azaharakis

I don't have experience with isomorphic apps, but I guess it's because document.body doesn't exist at compile time. Would it work if I do it like this instead?

````javascript
    var AppElement;

    componentDidMount: function() {
        ...
        (AppElement || document.body).appendChild(this.node);
        ...
    },
````

Comment (jackofseattle) [inline lib/components/Modal.js]: Seems there's a comment on every PR about something breaking when rendered on the server. Perhaps it's time for a simple unit test to catch these issues.  The dev community is pretty equally split on those that write isomorphic apps and those who don't, as such we can't expect everyone that works on this to know that what they're doing is going to break something.

Comment (jackofseattle) [inline lib/components/Modal.js]:
```js
var AppElement = ExecutionEnvironment.canUseDOM ? document.body : {appendChild: function() {}};
```

That will provide safe usage for server side.

evoyy: Sorry, I don't understand. If AppElement === {appendChild: function() {}}, surely AppElement.appendChild(this.node) will do nothing. So the modal will not be appended to the DOM.

jackofseattle: @evoyy - yes, it would do nothing. That's the point on the server side, we're just creating a mock so that it doesn't try to access the DOM unless there's something there.

evoyy: Ah, of course. I will amend my PR. Thanks.

jpserra:
Think you missed this https://github.com/rackt/react-modal/pull/108/files#diff-7388f7a885a57e359e87c12576ec0148R59

evoyy:
@jpserra Nice catch, thanks!

cameronprattedwards:
Thanks for fixing this. Would love to see this get merged. :smiley:

grsabreu:
Would love to get this on a release :D

jonathanp:
Same here, would love to have this merged!

feorex:
+1 for this PR!

claydiffrient:
@evoyy This looks good, but could you squash the four commits down to one.  It'll keep the git history a lot cleaner.  Another great thing to have would be some specs testing things out.

evoyy:
@claydiffrient Squashed and specs added.

PR #120, opened 2016-01-26
[fixed] Handle case when no tabbable element exists

evoyy:
Fixes https://github.com/rackt/react-modal/issues/44

### DIFF
--- a/specs/Modal.spec.js
+++ b/specs/Modal.spec.js
@@ -92,6 +92,14 @@ describe('Modal', function () {
     });
   });
 
+  it('handles case when child has no tabbable elements', function() {
+    var component = renderModal({isOpen: true}, 'hello');
+    assert.doesNotThrow(function() {
+      Simulate.keyDown(component.portal.refs.content, {key: "Tab", keyCode: 9, which: 9})
+    });
+    unmountModal();
+  });
+
   it('supports custom className', function() {
     var modal = renderModal({isOpen: true, className: 'myClass'});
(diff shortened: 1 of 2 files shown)


claydiffrient:
@evoyy This looks like a great change.  Can you add a spec that tests the functionality?  I generally like specs to accompany code changes.

evoyy:
@claydiffrient Spec added.

specs/Modal.spec.js
```diff
@@ -92,6 +92,12 @@ describe('Modal', function () {
     });
   });
 
+  it('handles case when child has no tabbable elements', function() {
```
claydiffrient: It apperas that you forgot to make an assertion in the spec. It just opens the modal, presses tab then closes it.

evoyy: That's intentional. The spec simply verifies that an exception is not thrown when a tab is pressed. Without the bugfix, the spec will fail.

evoyy: @claydiffrient I have updated the spec to use assert.doesNotThrow().

claydiffrient:
Okay, I checked out the code and took a look at the actual functionality this makes happen.  I'm slightly concerned because this fix allows focus to escape from the modal, which from an accessibility standpoint isn't ideal.  Granted I think most use cases of the modal will include a close button of some nature, which avoids this problem.  This is nice in that it avoids the errors being thrown when there is nothing to tab to when the user tabs like is currently on master.

I think this is good for now.  I want to think about maybe adding a way to switch between modal and non-modal in the future probably though.

PR #132, opened 2016-03-02
[Fixed] Prevent default behavior of ESC key

shunjikonishi:
Following problem is happened in my app.

- Close modal with ESC key
- HttpRequest which runs after closing modal is canceled
  - Because of ESC key


claydiffrient:
@shunjikonishi Looks simple enough, normally I'd prefer a spec for code changes, but a test in this case would be a little overkill perhaps.

Issue #71, opened 2015-09-25
Modals cannot be tested with React's TestUtil


baer:
react-modal is acting outside of the purview of any of React's test utilities.

``` js
renderedModal.getDomNode()
// => <noscript data-reactid=".0"></noscript>

React.scryRenderedDOMComponentsWithClass(renderedModal, "my-modal-class");
// => []
```

I see there are some work arounds that were developed circa React 0.12 but I can't figure out why that is necessary. If I was guessing I would say that this has to do with how you mount to the DOM to properly apply the overlay. If this is true, maybe a small wrapper around those helpers as a part of the release would be a helpful thing for us app devs who want to write lots of tests :)

https://github.com/rackt/react-modal/blob/master/specs/helper.js

dviramontes:
Having trouble testing this as well.
:+1: thanks for the pointer @baer !

jkimbo:
@baer I guess this is because the modal creates a "portal" by mounting the actual modal in a new div at the bottom of the page. You can see in the `Modal.js` component the render method is not actually returning anything: https://github.com/rackt/react-modal/blob/1ff6a02c689e9fdc03616c4ab82ec53317f14c69/lib/components/Modal.js#L74

To get around this you need to access the `.portal` property (https://github.com/rackt/react-modal/blob/1ff6a02c689e9fdc03616c4ab82ec53317f14c69/lib/components/Modal.js#L70) on the `renderedModal` to access the correct React tree to run the TestUtils on. So `
React.scryRenderedDOMComponentsWithClass(renderedModal.portal, "my-modal-class");` should work.

Hope that makes sense!

ArthurCid:
@jkimbo that sounded promising, unfortunately that didn't do the trick for me. I get an undefined on renderedModal.portal. Do you get the React tree on that ?

jkimbo:
@ArthurCid I've written up a passing test for you using mocha and jsdom: https://gist.github.com/jkimbo/872f42732e6c772bf8e1

gbrassey:
I'm still having trouble with this. I've tried `react-modal@0.6.1` w/ `react@0.14.2` and `react-modal@0.5.0` w/ `react@0.13.3`. Testing with karma, karma-chrome and mocha.

the `.portal` property exists and renders but it appears to have no children

```
import React from 'react';
import ReactDOM from 'react-dom';
import TestUtils from 'react-addons-test-utils';
import { expect } from 'chai';

import Modal from 'react-modal';

[...]

PR #136, opened 2016-03-14
Update README.md to include testing gotchas

codegoblin:
As this has been a source of confusion not only within the issues section of this repo, but also within my software consultancy, I am proposing the addition of the included 'testing gotchas' section to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,11 @@ var App = React.createClass({
 
 ReactDOM.render(<App/>, appElement);
 ```
+# Testing
+
+When using React Test Utils with this library, here are some things to keep in mind:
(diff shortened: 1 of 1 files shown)


cdidyk:
+1 This bit us in a big way and caused us to waste a lot of time before we thought to look at react-modal as the culprit and found [the relevant issue](https://github.com/reactjs/react-modal/issues/71) A helpful notice upfront that it requires some special handling would've been really useful.

claydiffrient:
@codegoblin It looks like you have 2 commits as part of this PR.  Could you merge it to one to keep history clean?  Then I'll merge it.

PR #141, opened 2016-03-26
Fix no tabbable focus

everdimension:
A follow up to https://github.com/reactjs/react-modal/pull/120.
This is a better way to handle the case when no tabbable elements are found within the modal's content.
The focus should be kept inside the modal the same way it's kept when there are some elements.


claydiffrient:
@everdimension Please squash the commits together then I'll merge it.

everdimension:
Sure, thanks the suggestion! Done

PR #128, opened 2016-02-19
[added] The ability to decide whether the modal should be closed when clicking the overlay area

dorsha:
This is an important ability since in some cases we don't want the modal to be closed when users are clicking outside.

PR #142, opened 2016-03-26
make all tests pass within es5 environment

everdimension:
All code for this library is written in `es5`, except for these two tests. So they fail when run in es5 environment. This PR fixes that.


claydiffrient:
This was fixed as part of #128, although I think I want to get all the environments switched over to ES2015+ soon.  In any case, thanks for this and your other PRs.

Issue #44, opened 2015-06-12
Error when pressing TAB


Semigradsky:
![image](https://cloud.githubusercontent.com/assets/1198848/8125088/2a320dd6-10eb-11e5-977c-1c6cedf4d18f.png)
`tabbable` array is empty.

My code:
```js
...
Modal.setAppElement(document.getElementsByTagName('body')[0]);
...
render() {
    return (
        <div>
            <Modal isOpen={true}>
                <div>some text</div>
            </Modal>
        </div>
    );
}
```

PR #145, opened 2016-03-29
[fixed] scopeTab no longer attempts to focus on undefined target

Zacqary:
When creating a modal with nothing tabbable inside it, I was getting `Cannot read property 'focus' of undefined` errors if I hit the Tab key (which was being listened to by another component). This PR will prevent that error from occurring.


claydiffrient:
@Zacqary Thanks for the PR... however, this issue was fixed with #120 and #141.  It'll be fixed in the next release we cut.  I'm tentatively planning on this Friday to do that.

Zacqary:
Aha, sorry about that. https://github.com/reactjs/react-modal/issues/44 had the error message in a screenshot so it didn't come up in a search.

claydiffrient:
No problem!

PR #148, opened 2016-04-06
[added] module for default style

tyrsius:
This new module allows `defaultStyles` to be overridden globally.
Test verifies that changed styles are used by the modal


tyrsius:
This doesn't actually work, since the built version of the library only exports the Modal component. This needs some more thought

claydiffrient:
Yeah, I was wondering how it would actually get the styles out there for use.  Otherwise, it looks like a good PR.

My thinking is maybe expose them via a `Modal.defaultStyles` or similar.

tyrsius:
I will do that. I've been working with ES6 too much lately and forgot named exports aren't a real thing in npm-land

tyrsius:
@claydiffrient ok, that should do it. I updated the README this time, too

claydiffrient:
Cool cool... I'm busy on other stuff right now, so I can't review it properly right now.  I'll look into it tonight and merge before I cut the next release.

PR #153, opened 2016-04-08
Add support for React v15.x in `peerDependencies`

jonathanp:
When installing this package in a project that uses React v15.x the following warning is shown:

`npm WARN react-modal@0.6.1 requires a peer of react@^0.14.0 but none was installed.`

This commit solves that.


claydiffrient:
Closed in favor of #152

Issue #51, opened 2015-06-29
How to focus on primary button when modal loads?


kevhou:
I can't seem to use jquery .focus() inside componentdidmount.

Someone please help!

olegshilov:
@kevhou try using [Refs](https://facebook.github.io/react/docs/more-about-refs.html#completing-the-example)

arasmussen:
I'm having a similar problem with not being able to .focus something inside componentDidMount. I'm using refs.

```javascript
var VIEWER_REF = 'VIEWER';
var PhotoViewer = React.createClass({
  componentDidMount: function() {
    this.refs[VIEWER_REF].getDOMNode().focus();
  },
  render: function() {
    return (
      <div tabIndex="0" ref={VIEWER_REF}>
        {contents}
      </div>
    );
  },
});
```

I've verified that this.refs[VIEWER_REF].getDOMNode().focus() is defined and being called. It worked before I put my PhotoViewer inside react-modal.

arasmussen:
Also worth noting that document.activeElement (currently focused element) after my PhotoViewer opens is the ReactModal__Content element, so it seems like it's stealing focus from me. Why does the modal need focus?

arasmussen:
Can we get a prop <Modal allowFocus={false} /> (default true) so that we can disable this functionality? I don't see why the Modal needs focus. I'm happy to manually pipe through ESC to changing isOpen to false.

arasmussen:
Also sorry for the quadruple post, but a temporary workaround is the following:

```javascript
componentDidMount: function() {
  setTimeout(function() {
    this.refs[MY_REF].getDOMNode().focus();
  }.bind(this), 0);
},
```

brandly:
In case anyone else comes across this, as of `v0.13`, [`getDOMNode()` is deprecated.](https://facebook.github.io/react/blog/2015/03/10/react-v0.13.html#new-features) The above example would be changed to this:

```js
componentDidMount: function() {
  setTimeout(function() {
    React.findDOMNode(this.refs[MY_REF]).focus();
  }.bind(this), 0);
},
```

SebT:
I'm having a similar problem, but I just can't get my ref.

```javascript

render() {
  return (
    <Modal>
        <input ref="input"/>
    </Modal>
}

componentDidMount() {
    console.log(this.refs.input); // undefined
}
```

[...]

Issue #50, opened 2015-06-26
How to perform action on modal open?


AndrewRayCode:
It seems that with the `isOpen` api, neither `componentDidMount` nor `componentDidUpdate` actually guarantee the modal is open, even when making sure props.isOpen is set. I'm not sure what causes the delay, but if I want to do something like focus an input on modal open, I can't do it because the modal isn't visible yet.

yhjor1212:
set it 'true'

DelvarWorld:
@yhjor1212 what? set what to true?

anatoliyarkhipov:
What did you mean by "delay"? From what moment it is delayed?

thesunny:
Thanks for putting this React Modal together. Really nice.

I think this and the issue on focus can be handled by having an onAfterOpen callback passed in as a prop. This way, after the modal is open, we can perform an action on open which includes the ability to set the focus.

My feeling is that it may be beneficial to remove the auto-focus since it's not necessary in a lot of cases and in cases where you want it, you can easily add focus through the onAfterOpen callback. In any case, if you do keep it, just a reminder to make sure the autofocus happens before the onAfterOpen callback or else any focus set in the callback would be lost.

Would also solve the following issue: https://github.com/rackt/react-modal/issues/51

claydiffrient:
Closing because #154 was released.

PR #154, opened 2016-04-08
[added] trigger onAfterOpen callback when available.

diasbruno:
Hi,

This patch would help to sync the refs to be accessed from the Component  (so, no need for a `setTimeout`). I've used the suggested name for the callback in #50 `onAfterOpen`.

This would help with #50 and #51.

Hope I followed the contribute guide.

### DIFF
--- a/lib/components/ModalPortal.js
+++ b/lib/components/ModalPortal.js
@@ -76,6 +76,10 @@ var ModalPortal = module.exports = React.createClass({
     focusManager.markForFocusLater();
     this.setState({isOpen: true}, function() {
       this.setState({afterOpen: true});
+
+      if (this.props.isOpen && this.props.onAfterOpen) {
+        this.props.onAfterOpen();
+      }
     }.bind(this));
   },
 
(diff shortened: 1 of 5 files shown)


lib/components/ModalPortal.js
```diff
@@ -76,6 +76,7 @@ var ModalPortal = module.exports = React.createClass({
     focusManager.markForFocusLater();
     this.setState({isOpen: true}, function() {
       this.setState({afterOpen: true});
+      (this.props.isOpen && this.props.onAfterOpen) && this.props.onAfterOpen();
```
victor-torres: Why not 
```
if(this.props.isOpen && this.props.onAfterOpen) { 
  this.props.onAfterOpen();
}
```
here?

diasbruno:
@claydiffrient just pinging to see if this helps with both mentioned issues and a little review in this PR. :)

claydiffrient:
@diasbruno Thanks for the ping.  The suggested change that @victor-torres made sounds good to me.  That will keep the code a bit more readable for those in the future.

diasbruno:
Awesome! I'll fix this.

lib/components/ModalPortal.js
diasbruno: Thanks @victor-torres! Fixed.

PR #164, opened 2016-04-19
[fixed] Make the non-minified dist build present again

### LINKED ISSUE #163: Non-minified build not present
### Summary:

All of the files in the `dist` directory are minified including `react-modal.js` instead of just `react-modal.min.js`

### Steps to reproduce:

 1. Run `./scripts/build`

### Expected behavior:

The `react-modal.js` should be transpiled, but not minified.  The `react-modal.min.js` should be both transpiled and minified.

claydiffrient:
Fixes #163 .

Changes proposed:
-  Adds back a non-minified build

Acceptance Checklist:
- [x] The commit message follows the guidlines in `CONTRIBUTING.md`.

This commit removes the `-p` flag from the call to webpack
so that the webpack config works as intended, creating two
files in the dist directory, one minified and one unminified.

After 3f323381b1e8c43fdbf172813b9e8a4ace9abe79 was merged the
build ended up only producing the production minified versions
for both files.  This commit restores the initial behavior
of a non-minified and a minified file.

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -1,2 +1,2 @@
 #!/bin/sh
-webpack --config webpack.dist.config.js -p
+webpack --config webpack.dist.config.js


diasbruno:
This is correct. +1

PR #176, opened 2016-05-17
[fixed] Make the modal portal render into body again

claydiffrient:
In c13fed9f3a1270a4ecfb6faf24355f2791948c00 behavior was changed
so that the modal portal ended up rendering into the app element.

In the default case where the appElement receives aria-hidden
when the modal is open, this meant that the modal itself was
also contained inside the aria-hidden element.  This limits the
ability for screenreaders to read the modal contents.

Server-side rendering should not be affected by this change
because componentDidMount is only fired on the client-side thus
having a reference to document.body there should be acceptable.

Changes proposed:
- Make the portal append to body rather than the app element

- Make sure that you are setting an app element (rather than using the default document.body) so that modal contents are not being aria-hidden

Acceptance Checklist:

### DIFF
--- a/lib/components/Modal.js
+++ b/lib/components/Modal.js
@@ -50,7 +50,7 @@ var Modal = React.createClass({
   componentDidMount: function() {
     this.node = document.createElement('div');
     this.node.className = 'ReactModalPortal';
-    AppElement.appendChild(this.node);
+    document.body.appendChild(this.node);
     this.renderPortal(this.props);
   },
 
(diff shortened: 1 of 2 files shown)


claydiffrient:
@diasbruno Would you mind looking this over and giving your thoughts before I merge and cut a new version?

diasbruno:
Sure!

diasbruno:
It's looks good. Good to go!

ricardocosta89:
Hello,

`README.md` should be updated with some urgency, because it's not clear how (or if possible) to have the modal rendered inside a specified DOM element.

Thanks!

PR #199, opened 2016-07-13
[fixed] Unable to tab through modal inputs (#198)

### LINKED ISSUE #198: Unable to tab through form elements
The content container is given a tabindex of -1 and it's not allowing me to tab through form elements. I've tried removing the attribute with jquery and readding the tabindexes to the inputs with jQuery, both to no avail. 

```
<div class="ReactModal__Content ReactModal__Content--after-open" tabindex="-1">
   <input type="text"/>
   <input type="text"/>
</div>
```

FuzzySockets:
Fixes #198.

  -Due to the 'visible' function recursively traversing up the DOM tree
  and checking the offset width and heights of all parents to determine
  visibility, any inputs contained within a ReactPortal with an offset
  height and width of 0/0 will be considered untabbable.

### DIFF
--- a/lib/helpers/tabbable.js
+++ b/lib/helpers/tabbable.js
@@ -20,7 +20,7 @@ function focusable(element, isTabIndexNotNaN) {
 }
 
 function hidden(el) {
-  return (el.offsetWidth <= 0 && el.offsetHeight <= 0) ||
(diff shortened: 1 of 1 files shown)


FuzzySockets:
Oops. That wasn't right.

PR #200, opened 2016-07-13
[fixed] Unable to tab through modal inputs (#198)

### LINKED ISSUE #198: Unable to tab through form elements

FuzzySockets:
Fixes #198.

  -Due to the 'visible' function recursively traversing up the DOM tree
  and checking the offset width and heights of all parents to determine
  visibility, any inputs contained within a ReactPortal with an offset
  height and width of 0/0 will be considered untabbable. I'm proposing 
  we make this visibility check an "OR"  since relying on a parents offset 
  height and width is unreliable since sometimes a parent element does 
  not wrap it's children depending on the applied styles.

### DIFF
--- a/lib/helpers/tabbable.js
+++ b/lib/helpers/tabbable.js
@@ -16,7 +16,7 @@ function focusable(element, isTabIndexNotNaN) {
     !element.disabled :
     "a" === nodeName ?
       element.href || isTabIndexNotNaN :
-      isTabIndexNotNaN) && visible(element);
+      isTabIndexNotNaN) || visible(element);
 }
 
 function hidden(el) {


FuzzySockets:
http://i.imgur.com/Eyc56Uk.png

FuzzySockets:
This logic isn't quite right, hmm.  Perhaps alter the hidden check to rely on other criteria besides offsetWidth and offsetHeight?

FuzzySockets:
I solved this in my local project by applying some additional styling. It's worth noting that if any element between a potential tabbable and body is 0px by 0px, in-modal tabbing will not function. It may be worth only checking visibility until you reach the root ModalPortal.

aradnom:
Ran into this issue myself, great find.

PR #201, opened 2016-07-21
Dont set body class if isOpen not change

khmelevskii:


claydiffrient:
@khmelevskii I'm afraid I'm going to need to revert this commit.  It seems that it actually breaks some tests :(  I failed to run them before merging.  It seems that  the biggest thing it does is fails to properly execute the appropriate things when the modal has an initial render that includes the modal being open.

I like the idea, but we'll need to revisit this with an alternative solution.

PR #205, opened 2016-08-01
[changed] Don't render div on dom when isOpen is false

CompuIves:
I noticed that when I wanted to animate the fadein of a modal it only worked for the first time. This was because the css animation of `ReactModal__Overlay--after-open` only works when a div is rendered for the first time. That's why I changed the logic to render nothing if there is no modal.
I'm aware that this can break a lot of things (like using a ref for `ReactModal`) for other people, so maybe we need to bump a major version if we want to have this.

Changes proposed:
- When not open, render `null` instead of `div`

When you want to use a `ref` on `ReactModal` and use it after it's rendered, you will have to place the call to the `ref` in a `setTimeout` so it's accessible after it's rendered.

Acceptance Checklist:

This allows users to add an animation for opening the modal.

### DIFF
--- a/specs/Modal.spec.js
+++ b/specs/Modal.spec.js
@@ -30,6 +30,11 @@ describe('State', () => {
     expect(ReactDOM.findDOMNode(mcontent(modal))).toNotExist();
   });
 
+  it('doesn\'t render the portal if modal is closed', () => {
+    const modal = renderModal({}, 'hello');
+    expect(ReactDOM.findDOMNode(modal.portal)).toNotExist();
+  })
+
   it('has default props', () => {
     const node = document.createElement('div');
     Modal.setAppElement(document.createElement('div'));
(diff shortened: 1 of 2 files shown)


diasbruno:
@claydiffrient any thoughts? looks good.

claydiffrient:
@CompuIves Sorry for the delay on this.  Could you change the commit message from [fixed] to [changed] ?  I think that will make it easier to indicate that it's more of a breaking change vs. a simple bug fix. I believe this is a valuable change.  I'd like to get it in for v2.

CompuIves:
No problem! I changed it.

diasbruno:
Hi @CompuIves, we have released `v2.0.0`. This will need a rebase.

CompuIves:
Okay! Will do it today.

CompuIves:
There were a lot of changes, I decided to just recreate the changes based on new `master`. 

I wasn't able to reproduce the rendering of `div` instead of `null` in the tests anymore, maybe I'm doing something wrong or something has changed.

diasbruno:
Awesome, @CompuIves! 
Perhaps after close the modal you can check if `modal.portal == null`.

CompuIves:
Ah, that works, thanks! Added test.

diasbruno:
yeah, that is correct.

diasbruno:
You can squash the commits and append `[chore]` at the begging of the commit message.

CompuIves:
Done!

PR #208, opened 2016-08-08
[added] ability to change default 'ReactModalPortal' class

Leimi:
Changes proposed:
- new `portalClassName` prop to change the *ReactModalPortal* class.

~~Upgrade Path (for changed or removed APIs):~~

Acceptance Checklist:
- [x] ~~If the commit message has [changed] or [removed], there is an upgrade path above.~~

_____________

Thanks for this project! 

What do you think about adding this?

I think this is useful because:
  - code can be cleaner when also using custom content and overlay classnames. If using *BEM* classes, it makes more sense to have a DOM like `.CustomModal > (.CustomModal-content + .CustomModal-overlay)` and not like `.ReactModalPortal > (.CustomModal-content + .CustomModal-overlay)`.
  - my current use case is opening a modal from a content-script in an browser extension. I want to be able to target the portal DOM element precisely in order to prevent conflicts with the page css.

Few questions on my side:

- I went with "portalClassName" but I'm not sure if this is really speaking to everyone.
- I added a brief sentence about the change in the README but didn't add anything in the examples. I figured the readme change would be enough, but let me know if you don't think so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ you can pass `className` and `overlayClassName` props to the Modal.  If you do
 this then none of the default styles will apply and you will have full control
 over styling via CSS.
 
+You can also pass a `portalClassName` to change the wrapper's class (*ReactModalPortal*).
+This doesn't affect styling as no styles are applied to this element by default.
 
 ### Overriding styles globally
 The default styles above are available on `Modal.defaultStyles`. Changes to this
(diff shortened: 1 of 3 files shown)


claydiffrient:
Looks good to me and I see no reason this can't go through.

PR #197, opened 2016-07-12
[fixed] Fixes issue#131 -  Drag doesn't close modal

abhijitkane:
Fixes #[131].

Changes proposed:
- Mouse drag from the content window to the overlay doesn't close the modal

Acceptance Checklist:
- [x] All commits have been squashed to one.
- [x] The commit message follows the guidelines in `CONTRIBUTING.md`.
- [ ] Documentation (README.md) and examples have been updated as needed.
- [x] If this is a code change, a spec testing the functionality has been added.
- [ ] If the commit message has [changed] or [removed], there is an upgrade path above.

Issue #131, opened 2016-03-01
Dragging outside of the modal closes the modal


megapctr:
I have a modal with some text in it. When I select the text and accidentally move my mouse outside the dialog, releasing the mouse button closes the dialog.

How to reproduce:
  1. Click inside the modal
  2. Drag outside the modal
  3. Release the mouse button

What happens:
  - The modal closes

claydiffrient:
This was closed by #217

PR #217, opened 2016-08-30
Close modal when mouseDown and MouseUp happen only on the overlay

d-mon-:
Changes proposed:
- If the user perform a mouse down and up only on the overlay: close the modal
- If the user perform a mouse down on the overlay, then a mouse up on the content: do nothing
- If the user perform a mouse down on the content, then a mouse up on the overlay: do nothing

Moreover, I tried to avoid the use of stopPropagation.

And it should also fix the [issue#131](https://github.com/reactjs/react-modal/issues/131) resolved by @abhijitkane in his [pull request](https://github.com/reactjs/react-modal/pull/197) by handling this problem with mouse down events only and stopPropagation.

Acceptance Checklist:

finally, I'm not really sure if it's a bug fix or not... :)

### DIFF
--- a/lib/components/ModalPortal.js
+++ b/lib/components/ModalPortal.js
@@ -21,6 +21,7 @@ var CLASS_NAMES = {
 var ModalPortal = module.exports = React.createClass({
 
   displayName: 'ModalPortal',
+  shouldClose: null,
 
   getDefaultProps: function() {
     return {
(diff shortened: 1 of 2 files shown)


Robinnnnn:
@claydiffrient I see that this was merged 16 days ago but the latest release was `1.4.0` back in June. is it possible to pull this fix via npm?

claydiffrient:
My plan was to include it in our 2.0 release, but that's being held up by #183 right now.  I should be able to add it into a minor/patch release later today, to get us by until then.

PR #222, opened 2016-09-12
[fixed] Don't steal focus from a descendent when rendering

### LINKED ISSUE #220: focus/select text in input
I have problem with focusing and then selecting text in input in modal. Normal out of modal this work - 
```
render()
return(
<input className={'numpadInput'} onChange={this.handleChange} value={this.state.value} ref="nInput" type="text"/>)
)
componentDidMount(){
const input = ReactDOM.findDOMNode(this.refs.nInput);
    input.focus();
    input.select();
}
```
But in modal don't. Why?

conlanpatrek:
Fixes #220.

Changes proposed:
- Before focusing on the content div, check to see that the currently focused element isn't a descendent of the content div, to respect any focusing the application may have already done.

Acceptance Checklist:

### DIFF
--- a/lib/components/ModalPortal.js
+++ b/lib/components/ModalPortal.js
@@ -98,7 +98,10 @@ var ModalPortal = module.exports = React.createClass({
   },
 
   focusContent: function() {
-    this.refs.content.focus();
+    // Don't steal focus from inner elements
+    if (!this.contentHasFocus()) {
+      this.refs.content.focus();
+    }
   },
 
   closeWithTimeout: function() {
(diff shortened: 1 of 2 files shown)


claydiffrient:
I'm confused by this.  The modal contents, before being opened doesn't exist in the DOM, so nothing inside it should be able to be focused prior to the modal opening.

If you need to set focus to something after the modal has opened, you can you use the `onAfterOpen` function prop to set focus to something other than the content div immediately after opening.

conlanpatrek:
React life cycle events and ref callbacks are called on children before they are called on parents. When `componentDidUpdate` is called on `ModalPortal`, it's already been called on all of it's descendants, so there plenty of opportunity for a child to have already given focus to some element in the modal. Right now though, that focus is being reassigned to the content div immediately afterward.

You _could_ set focus in the `onAfterOpen` hook, but that would require you to traverse down from the component that opened the modal, coupling all of the components you pass through (especially problematic if your focusable element is deeply nested). Devs also already manage focus with `componentDidMount/Update` and `ref` callbacks. It seems strange that they'd need to implement a second set of focus logic to account for their component appearing in a modal.

claydiffrient:
Good points.  I don't see any reason that we shouldn't implement this.

PR #230, opened 2016-09-23
[added] Apply custom data- and aria- props to content div

Munter:
Changes proposed:
- Apply `aria-*` and `data-*` props to the modal content div

Acceptance Checklist:

### DIFF
--- a/lib/components/ModalPortal.js
+++ b/lib/components/ModalPortal.js
@@ -1,4 +1,5 @@
 var React = require('react');
+var HTMLDOMPropertyConfig = require('react/lib/HTMLDOMPropertyConfig');
 var div = React.DOM.div;
 var focusManager = require('../helpers/focusManager');
 var scopeTab = require('../helpers/scopeTab');
@@ -17,6 +18,18 @@ var CLASS_NAMES = {
     beforeClose: 'ReactModal__Content--before-close'
   }
 };
+
+function getCustomProps(props) {
+  var customProps = {};
+
+  Object.keys(props).forEach(function (key) {
+    if (HTMLDOMPropertyConfig.isCustomAttribute(key)) {
+      customProps[key] = props[key];
+    }
+  });
+
+  return customProps;
+}
 
 var ModalPortal = module.exports = React.createClass({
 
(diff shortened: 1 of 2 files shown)


diasbruno:
This looks good for aria attributes.

lib/components/ModalPortal.js
```diff
@@ -18,6 +19,18 @@ var CLASS_NAMES = {
   }
 };
 
+function getCustomProps(props) {
```
diasbruno: Maybe move this function to a helper?

afercia:
It would be great to have the ability to add ARIA attributes on the modal, this would be necessary for example when the modal content is mainly text:

screen shot 2017-05-11 at 16 58 53

The text inside the modal won't be announced because screen readers enter in focus mode when `role=dialog` is set., Using `aria-describedby` to target the text and optionally `aria-labelledby` to target the visible heading would solve the accessibility issue.

diasbruno:
Hi @Munter, we have `v2.0.0` available now. We could make this PR to allow the user to pass `data-*` attributes and later we can think about adding `aria-*`. Any thoughts?

Munter:
@diasbruno Thanks for the heads up. I switched to react-a11y-modal because this was a blocker to me. I don't have the bandwidth to port this to v2

diasbruno:
@Munter no problem. Thank you.

diasbruno:
@Munter Do you mind if I slice your PR and add to `react-modal`?

Munter:
Take and modify whatever you need from this. It would be a fine addition to the library to get those features in

WickyNilliams:
Is this no longer supported?

diasbruno:
@WickyNilliams `aria-` attributes is available. See [https://github.com/reactjs/react-modal#additional-aria-attributes](https://github.com/reactjs/react-modal#additional-aria-attributes).

WickyNilliams:
ah... i specifically wanted data-* attributes, as hooks for targeting elements in tests

diasbruno:
One option is place those attributes in a wrapper on the content `<div />` or something else.

```jsx
<Modal ...>
    <div data-x=... >...</div>
</Modal>
```

diasbruno:
@WickyNilliams Can you open a new issue to make it easy to track?

WickyNilliams:
Yeah that's what I did eventually, but not ideal.

Sure thing! I'll make one now

Issue #214, opened 2016-08-19
Add `onAfterClose` property


halt-hammerzeit:
A proper modal window resets itself upon closing.
It can be reset immediately when `isOpen` is set to `false` but that introduces glitches due to resetting and repainting the modal window before closing animation is played.
An `onAfterClose` handler run in `componentWillUnmount` would resolve this issue.
I can make a PR if you are willing to accept this feature.

brentluna:
This would be awesome.

megapctr:
While this prop would be very helpful, my case was simple enough to implement it manually:
```
    onRequestClose = () => {
        this.props.closeModal()
        setTimeout(
            this.props.resetModal,
            this.props.closeTimeoutMS
        )
    }
```

halt-hammerzeit:
@megapctr oh, yeah, didn't think about this kind of workaround, that will do (though that's just a workaround and doesn't fix the issue)

halt-hammerzeit:
@megapctr And also this workaround won't work for programmatically closing the modal since it will either have to call `this.refs.modal.close()` (which is non-React way) or have to be upgraded to duplicate [`ModalPortal`](https://github.com/reactjs/react-modal/blob/master/lib/components/ModalPortal.js#L54) functionality:

```js
  componentWillReceiveProps: function(newProps) {
    // Focus only needs to be set once when the modal is being opened
    if (this.props.isOpen && !newProps.isOpen) {
        setTimeout(
            this.props.resetModal,
            this.props.closeTimeoutMS
        )
    }
  },
```

halt-hammerzeit:
I found another way: simply componentWillUnmount() in the modal contents component

PR #239, opened 2016-10-05
[added] [changed] made all classes customizable with a prop for each;…

rfreitas:
Changes proposed:
- All class names can be customised, this is helpful for people who want to name space their classes.
- Derivate classes, `--after-open` and `--before-close`, use their base classes as a constructor.
- Specs: `renderModal` helper function now re renders the modal if it already exists, this made a commented spec pass, so it has been included; `unmountModal` is called after each spec, so not needed calls were removed.

`--after-open` and `--before-close` derivate classes are now constructed from the passed prop base class name.
CSS style which is based on the old behaviour needs to update.

Acceptance Checklist:

UPDATE: It seems I have sneaked in an extra feature mentioned in issue #214. If the feature is wanted I can write a spec for it, if not I can remove it from the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,12 @@ At this time, media queries will need to be handled by the consumer.
 If you prefer not to use inline styles or are unable to do so in your project,
 you can pass `className` and `overlayClassName` props to the Modal.  If you do
 this then none of the default styles will apply and you will have full control
-over styling via CSS.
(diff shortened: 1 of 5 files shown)


diasbruno:
Hi @rfreitas, thank you so much for taking time working on this. Unfortunately, this has been superseded by a12246e.

I'll close this PR, but please, feel free to open a new one if something is can be improved.
Thanks.

PR #243, opened 2016-10-08
[added] Add contentLabel prop to put aria-label on modal content

### LINKED ISSUE #236: The dialog must be properly labeled
Hello, first of all thanks very much for the excellent work done so far, especially the focus manager. 

I'd recommend to consider that one very important thing for accessibility is currently missing, and it's a proper label on the dialog main `<div>` element. It's especially important now, with the introduction of the dialog role in #223.

Please consider implementations across screen readers differ, not all of them behave the same way. For example, testing with VoiceOver on Safari (Mac), it automatically reads out the main heading in the content as soon as the modal opens (at least for me).

By the way, testing with NVDA on Firefox (Windows) nothing is announced as the modal opens. That's understandable since the main div gets the focus and there's really nothing that can be announced there. Worth nothing this is the most standard behaviour, while VoiceOver has a different interaction model compared with other screen readers.

There are two possible, well known, best practices: use `aria-labelledby` or `aria-label`. By the way, I guess the real issue is that there's no guarantee authors will actually use a "title" in their content.

I'd suggest to consider to make the title required and than use that as target for `aria-labelledby`.
Maybe, the title could be optionally hidden and then used just for the `aria-label` attribute. For sure there are some things to carefully consider, but labelling the dialog is really important since with some screen readers there is just silence when the modal opens.

Some references:

https://www.w3.org/TR/wai-aria/roles#dialog
> Authors SHOULD provide a dialog label.

https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal
> The dialog box title is provided by either the aria-label or the aria-labelledby property.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_dialog_role
> The dialog must be properly labeled

Advanced ARIA Tip #2: Accessible modal dialogs
https://www.marcozehe.de/2015/02/05/advanced-aria-tip-2-accessible-modal-dialogs/

claydiffrient:
Fixes #236.

Changes proposed:
- Add a aria-label to the modal content

 - Existing modals should add a `contentLabel` prop describing the modal content

Acceptance Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ npm install --save react-modal
   onRequestClose={requestCloseFn}
   closeTimeoutMS={n}
   style={customStyle}
+  contentLabel="Modal"
 >
   <h1>Modal Content</h1>
   <p>Etc.</p>
@@ -136,7 +137,9 @@ var App = React.createClass({
           isOpen={this.state.modalIsOpen}
           onAfterOpen={this.afterOpenModal}
           onRequestClose={this.closeModal}
-          style={customStyles} >
+          style={customStyles}
+          contentLabel="Example Modal"
+        >
 
           <h2 ref="subtitle">Hello</h2>
           <button onClick={this.closeModal}>close</button>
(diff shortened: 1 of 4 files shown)


claydiffrient:
@diasbruno / @afercia would you mind reviewing this?

diasbruno:
@claydiffrient Looks good! Ship it!

diasbruno:
hmmmm, #230 would let people to choose what they want for data and aria attributes.

claydiffrient:
@diasbruno It doesn't look like #230 support all props just custom props.  I think this is acceptable to make sure that the aria-label can be applied.  I think once the merge conflicts are resolved on #230 then we can merge it as well.

PR #253, opened 2016-10-13
[fixed] respect closeTimeoutMS during unmount

### LINKED ISSUE #17: Unmounting and animation
When the modal is unmounted, it will abruptly close, not waiting for any animations to finish. Sometimes it is desired for the modal to graciously close with animations instead.

What I'm not too sure yet is how this should be handled, should it be a prop that is sent to tell that the modal should be closed on unmount? Should that be the normal behavior?

What do you guys think? Have you ever needed this in your apps?

mgh:
Fixes #17.

Changes proposed:
- wait before unmounting the portal if `closeTimeoutMS` is set
- full details in commit message

- The upgrade path is to undo any workarounds or hacks (e.g. manually triggering the close animation before unmounting the `Modal`), though due to the lack of activity on the referenced issue, I'm guessing most users won't be affected?

Acceptance Checklist:

### DIFF
--- a/lib/components/ModalPortal.js
+++ b/lib/components/ModalPortal.js
@@ -132,8 +132,9 @@ export default class ModalPortal extends Component {
   }
 
   closeWithTimeout () {
-    this.setState({ beforeClose: true }, () => {
-      this.closeTimer = setTimeout(this.closeWithoutTimeout, this.props.closeTimeoutMS);
+    const closesAt = Date.now() + this.props.closeTimeoutMS;
+    this.setState({ beforeClose: true, closesAt }, () => {
+      this.closeTimer = setTimeout(this.closeWithoutTimeout, this.state.closesAt - Date.now());
     });
   }
 
@@ -142,7 +143,8 @@ export default class ModalPortal extends Component {
       beforeClose: false,
       isOpen: false,
       afterOpen: false,
-    }, () => this.afterClose());
+      closesAt: null
+    }, this.afterClose);
   }
 
   handleKeyDown = (event) => {
@@ -182,7 +184,7 @@ export default class ModalPortal extends Component {
   }
 
   shouldBeClosed () {
-    return !this.props.isOpen && !this.state.beforeClose;
+    return !this.state.isOpen && !this.state.beforeClose;
   }
 
   contentHasFocus () {
--- a/lib/components/Modal.js
+++ b/lib/components/Modal.js
@@ -105,6 +105,24 @@ export default class Modal extends Component {
       ariaAppHider.show(this.props.appElement);
     }
 
+    const state = this.portal.state;
+    const now = Date.now();
+    const closesAt = state.isOpen && this.props.closeTimeoutMS
+      && (state.closesAt
+        || now + this.props.closeTimeoutMS);
+
+    if (closesAt) {
+      if (!state.beforeClose) {
+        this.portal.closeWithTimeout();
+      }
+
+      setTimeout(this.removePortal.bind(this), closesAt - now);
+    } else {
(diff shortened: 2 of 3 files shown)


lib/components/ModalPortal.js
```diff
@@ -166,7 +168,7 @@ var ModalPortal = module.exports = React.createClass({
   },
 
   shouldBeClosed: function() {
-    return !this.props.isOpen && !this.state.beforeClose;
+    return !this.state.isOpen && !this.state.beforeClose;
```
mgh: if the `Modal` is unmounting, it won't have a chance to update the `isOpen` prop to `false`; however, the internal `setState` call will update `state.isOpen` correctly.

claydiffrient:
@mgh This looks good from a code perspective, but it'd be great to have a spec in place.  I think we could do an async spec that verified that the time the callback is called is within some small margin of error from the timeout value specified.

mgh:
@claydiffrient sounds good.  I've updated the test specs to include an async spec as you proposed.  The new test queries the DOM after the modal is unmounted to ensure the portal still exists until the `closeTimeoutMS` expires, so I had to update the other tests to ensure other modals weren't still lying around.  (I noticed a related test for `closeTimeoutMS` that was commented out because it didn't work, so I fixed that as well.)

diasbruno:
@mgh awesome! Is this action cancelable? (Just a question).

mgh:
@diasbruno depends on what you mean by cancelable =) the original behavior of closing and then immediately reopening a modal should be unaffected (the first close is canceled and the portal is preserved).  Unmounting then remounting the same Modal will not preserve the same portal though, as React instantiates a new Modal instance for each mount and each instance has its own portal.  That behavior, though, is the same before and after this commit.

mgh:
@diasbruno I rebased.  Two things to note:

1. Original change (https://github.com/reactjs/react-modal/commit/ebec6386156fe7fbe57f3365cfdce6d26fb4c962#diff-9e8d9aa8af75942f5423a4b030c5b78aR145):
    ```
    this.setState({
       ...                                                   
    }, this.afterClose);          →        }, () => this.afterClose());
    ```
    
    However, this same commit redefined `afterClose`:
    
    ```
    afterClose = () => {
    ```
    
    So I kept the original `}, this.afterClose)` since the extra function wrapper does not appear     necessary.

2. A new test did not unmount the modal, breaking my test which expects no modals to be left-over from previous tests.  This same problem may affect tests in the future, and it's hard to debug since adding a new test can break my (unrelated) test.  Thus, in addition to fixing the test in question, I also added:

    ```
    afterEach('check if test cleaned up rendered modals', () => { ... });
    ```

    which will now correctly identify tests that don't clean up:

    ```
    ...
    ✔ give back focus to previous element or modal.
    ✖ "after each" hook: check if test cleaned up rendered modals for "give back focus to previous element or modal."

    Finished in 1.446 secs / 0.072 secs

    SUMMARY:
    ✔ 11 tests completed
    ℹ 2 tests skipped
    ✖ 1 test failed

    FAILED TESTS:
      Modal
        ✖ "after each" hook: check if test cleaned up rendered modals for "give back focus to previous element or modal."
   ```

   The `afterEach` hook could try automatically unmounting all modals first to make test writing easier.

diasbruno:
Yeah, I've added this wrapper on `afterClose` (my bad) and forgot to remove. I add this because It was not firing cause it was a class method and not instance method or something.

Really like the `afterEach`, it is the best place to do some clean up for the next test. I believe we should add a function to force unmount any remain modal after a test.

diasbruno:
Tested and working.

PR #265, opened 2016-11-20
[added]Provide a hook to enable infinite scroll in modal

kenny-hibino:
I was trying to implement a infinite scroll feature within the modal using this library. I soon discovered that I had to fork the repo and make changes to let me provide a callback on scroll event. I was successful in implementing the feature after making the changes and thought it would be useful to have the same hook in the original library. 

It would be useful when you are trying to implement a modal to display a list of users (followers, people liked a post, etc..) and need to fetch more resource as you scroll down inside the modal. 

Changes proposed:
- Add two props `onEndReached` and `onEndReachedThreshold` to let user provide a callback function when bottom of the modal is reached.
- This provides a hook to enable infinite scroll feature inside the modal

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,29 @@ You can use this to remove scrolling on the the body while the modal is open.
 }
 ```
 
+### Provide an infinite scroll hook
+You can provide a callback function to be invoked when a user scroll down to
+the end of the modal. Provide `onEndReached` (function) and `onEndReachedThreshold` (number) via props.
+`onEndReached` will be called when modal has been scrolled within onEndReachedThreshold
+of the bottom.
+`onEndReachedThreshold` is threshold in pixel for calling `onEndReached`, defaults to 20.
+
+```jsx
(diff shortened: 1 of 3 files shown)


diasbruno:
@kenny-hibino, this feature looks really good, but maybe it can be a separate module that can be used within the modal.

```javascript
<Modal { modal props }>
  <InfinityScroll { props for the infinity scroll } />
</Modal>
```

kenny-hibino:
@diasbruno Thank you for replying! I will try to see if I can extract this out into a separate module as you suggested 👍

diasbruno:
@kenny-hibino awesome! It'll be an awesome component because people could use it anywhere.

diasbruno:
@claydiffrient This issue can be closed.
@kenny-hibino Remember that you still can publish your own version with this feature...and/or if you implement a infinity scroll component, remember to share it! :)

PR #266, opened 2016-11-22
[fixed] hidden check to read `getClientBoundingRect` instead of `offset`

damonbauer:
Fixes #224.

Changes proposed:
- Includes code change from #224
- Adds tests to that were requested by @claydiffrient: https://github.com/reactjs/react-modal/pull/224#issuecomment-252405710
- Adds example of tabbable element inside of an `inline` element

Upgrade Path (for changed or removed APIs): *none*

Acceptance Checklist:
- [x] All commits have been squashed to one.
- [x] The commit message follows the guidelines in `CONTRIBUTING.md`.
- [x] Documentation (README.md) and examples have been updated as needed.
- [x] If this is a code change, a spec testing the functionality has been added.
- [x] If the commit message has [changed] or [removed], there is an upgrade path above.

PR #224, opened 2016-09-15
[fixed] hidden check was breaking tabbing through forms

evanhobbs:
Fixes #211 .

This is regarding: https://github.com/reactjs/react-modal/issues/211 and credit goes to @nickb1080

Changes proposed:
- use getBounding rect to determine if an element if visible or not

Upgrade Path (for changed or removed APIs):
- no breaking changes

Acceptance Checklist:
- [x] All commits have been squashed to one.
- [x] The commit message follows the guidelines in `CONTRIBUTING.md`.
- [x] Documentation (README.md) and examples have been updated as needed.
- [x] If this is a code change, a spec testing the functionality has been added.
- [x] If the commit message has [changed] or [removed], there is an upgrade path above.

PR #271, opened 2016-12-07
[fixed] Respect overflow css property when determining whether or not a tabbable node is hidden

### LINKED ISSUE #211: I can't navigate trough form elements using tab button
### Steps to reproduce:
1. add some form elements inside a popup
2. try to navigate with tab button
### Expected behavior:

focus should move trought the form elements

conlanpatrek:
Fixes #211.

Changes proposed:
When checking whether or not a node is hidden, we traverse up through the body element and say "Hey, if this element is 0x0, then it is hidden. Don't allow tabbing to its descendant." But if I, for example, position an element absolutely from its parent, theres a good chance that parent may collapse to 0x0. And if that parent has an overflow of 'visible', my input element is still visible and should be tabbable.

In our case it was `.ReactModalPortal` that was collapsing down to 0x0 and failing the `tabbable` check because it was considered `hidden`, even though our modal was visible. In this case, tabbing within a modal had no effect, as everything was descendent from the portal. My guess is most of the struggles people are having with tabbing come from a similar setup. **As a work-around you can fix tabbing by setting the width or height of your collapsed parent element(s) to 1px**

I've fixed this by changing the check from `hidden` to `hidesContents`, and altered it to respect the `overflow` property when the node has contents.

I've also gone ahead and added testing for the entire `tabbable` helper.

Upgrade Path (for changed or removed APIs): n/a

Acceptance Checklist:

### DIFF
--- a/specs/Modal.helpers.spec.js
+++ b/specs/Modal.helpers.spec.js
@@ -1,0 +1,110 @@
+/* eslint-env mocha */
+import "should";
+import tabbable from "../src/helpers/tabbable";
+import "sinon";
+
+export default () => {
+  describe("tabbable", () => {
+    describe("without tabbable descendents", () => {
+      it("returns an empty array", () => {
+        const elem = document.createElement("div");
+        tabbable(elem).should.deepEqual([]);
+      });
+    });
+
+    describe("with tabbable descendents", () => {
+      let elem;
+      beforeEach(() => {
+        elem = document.createElement("div");
+        document.body.appendChild(elem);
+      });
+
+      afterEach(() => {
+        document.body.removeChild(elem);
+      });
+
+      it("includes descendent tabbable inputs", () => {
+        const input = document.createElement("input");
+        elem.appendChild(input);
+        tabbable(elem).should.containEql(input);
+      });
+
+      it("includes tabbable non-input elements", () => {
+        const div = document.createElement("div");
+        div.tabIndex = 1;
+        elem.appendChild(div);
+        tabbable(elem).should.containEql(div);
+      });
+
+      it("includes links with an href", () => {
+        const a = document.createElement("a");
+        a.href = "foobar";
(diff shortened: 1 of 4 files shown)


conlanpatrek:
One thing I wasn't quite sure of... It looks like tabbable will traverse detached nodes up to their root and potentially return true... Is that intentional?

kadikraman:
@conlanpatrek @diasbruno Any plans to merge this?

mandazi:
This PR fixes the tab issue I was having mentioned in #211

conlanpatrek:
@diasbruno I have rebased this. Was forced to shift some stuff around for the lint constraints. Take a look and let me know if you have any questions.

specs/Modal.helpers.spec.js
```diff
@@ -0,0 +1,110 @@
...
+        const input = document.createElement("input");
+        elem.style.width = "0";
+        elem.style.height = "0";
+        elem.style.overflow = "hidden";
+        elem.appendChild(input);
+        tabbable(elem).should.not.containEql(input);
+      });
+
+      it("includes inputs visible because of overflow == visible", () => {
+        var input = document.createElement("input");
```
conlanpatrek: oop. missed a lonely var. will fix.

diasbruno:
Looks good! @mandazi Can you confirm the fix? If it is ok, you can approve and we merge this.

mandazi:
@diasbruno Confirmed. It works!

diasbruno:
Thank you @mandazi and @conlanpatrek!

diasbruno:
Released `v3.1.3`.

conlanpatrek:
@diasbruno It might be worth checking if this solves the "me toos" presented in the comment threads of PR #224 and #266

diasbruno:
Sure, I'll try to check those PRs later.

PR #276, opened 2016-12-13
Move documentation site to GitBook

claydiffrient:
I wanted to push this up to start getting some feedback on it.

This would remove the custom documentation page that I was working on a while back, replacing it instead with GitBook (similar to how the Redux documentation is done).

In thinking about this a bit more now, perhaps something like React Storybook might be a better fit for a repository lighter on words an heavier on examples.

In any case, if you'd like to pull this down and check it out simply `npm install` or `yarn`, then run `npm run docs-dev` and that should get things going for you so they can be looked at.

Please let me know your thoughts.  It's definitely incomplete at this moment, but is getting closer everyday :)


claydiffrient:
@diasbruno I'd love your input here :)

diasbruno:
Awesome!

diasbruno:
@claydiffrient hope today I can give a help on this.

claydiffrient:
@diasbruno Sweet that'd be great.  I kinda decided I like react-storybook a bit better.  It keeps us from having to keep the runnable code external in CodePen.  I'll push up another WIP with what I've got so far with it soon.

claydiffrient:
@diasbruno Okay... so that being said... there appears to be a problem displaying the source properly for more complex components.  React storybook seems to be great to demo a single component in various configurations.  It doesn't seem great for showing a component that essentially needs to be wrapped such as react-modal.

Maybe we should just keep going down the GitBook path for now.

diasbruno:
I'm writing a small app to show how to use the modal events and docs.

[diasbruno/react-modal/chore/docs](https://github.com/reactjs/react-modal/commit/18611edc1d4b21830f964f2710ab90ef1716c503)
[codepen.io/diasbruno](http://codepen.io/diasbruno/pen/rWbYQQ)

PR #281, opened 2016-12-16
[added] Add Overlay Extras

cema-sp:
I have a use case where some text should be placed in bottom right corner of screen just below modal.
To have that text positioned relative to modal one can place it into modal itself, but to position text relative to screen it should be inside overlay.
Extras allows users to have some additional components inside modal overlay.

Changes proposed:
- Add `overlayExtras` prop allowing users to have additional elements inside modal overlay (e.g. info message, link, copyright)
- Add `overlayExtrasClassName` prop allowing to style extras div
- Add `styles.extras` prop (along with default values) allowing to style extras div
- Add extras example to basic examples

Acceptance Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Example:
 ```
 
 ## Styles
-Styles are passed as an object with 2 keys, 'overlay' and 'content' like so
+Styles are passed as an object with 3 keys, 'overlay', 'content', 'extras' like so
 ```js
 {
   overlay : {
@@ -58,7 +58,11 @@ Styles are passed as an object with 2 keys, 'overlay' and 'content' like so
     borderRadius               : '4px',
     outline                    : 'none',
     padding                    : '20px'
-
+  },
+  extras: {
+    position : 'absolute',
+    bottom   : '10px',
+    right    : '10px'
   }
 }
 ```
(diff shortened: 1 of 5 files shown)


claydiffrient:
Thanks for the PR.  I'm not sure about the API for this.  I'd like to avoid putting too many more props on the modal itself.  What if it was a separate `<ModalExtras />` component the modal could render instead?  Actually I'm not sure I like that solution any better either.

@diasbruno Thoughts?

diasbruno:
@cema-sp @claydiffrient maybe we could give to `MyOverlay` > `Modal`. It would require, at least, to render `this.props.children` to render the modal.

Maybe something that we could just...
```javascript
class MyOverlay extends ModalOverlay
```
```javascript
<Modal {...} overlay=<MyOverlay /> />
```

claydiffrient:
That actually sounds really good to me.  Just have custom overlay components that can be sent in to the Modal.

cema-sp:
@claydiffrient @diasbruno thank you for answers. I think solution with custom overlay overriding default would be nice and I'll try to do it.
Do you think we should extract default `ModalOverlay` to separate component (to be able to inherit from it)?

diasbruno:
@cema-sp yes, that would make it easier to `extend`.

johnwiseheart:
Do you reckon we can get this merged? Would love to use these features.

claydiffrient:
@johnwiseheart As this PR stands it isn't ready to be merged.  There needs to be a larger discussion around the API for handling this.

diasbruno:
I'm closing this PR in favor of #319. 
Thank you, @cema-sp.

PR #295, opened 2017-01-01
Modify the sample code to es2015 syntax in README.md

abouthiroppy:
Changes proposed:
We use `const` in README.md's sample.

Acceptance Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,9 +122,9 @@ You can use this to remove scrolling on the the body while the modal is open.
 Inside an app:
 
 ```js
-var React = require('react');
-var ReactDOM = require('react-dom');
-var Modal = require('react-modal');
+import React from 'react';
+import ReactDOM from 'react-dom';
+import Modal from 'react-modal';
 
 
 /*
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -122,9 +122,9 @@ You can use this to remove scrolling on the the body while the modal is open.
 Inside an app:
 
 ```js
-var React = require('react');
-var ReactDOM = require('react-dom');
-var Modal = require('react-modal');
+const React = require('react');
```
claydiffrient: Should we go ahead and do `import React from 'react';` here?

README.md
```diff
@@ -155,7 +155,7 @@ const customStyles = {
 };
 
 
-var App = React.createClass({
+const App = React.createClass({
```
claydiffrient: Perhaps convert this over to ES2015 class syntax as well?

abouthiroppy: OK, so I'll change es2015 syntax :)

abouthiroppy:
@claydiffrient Updated.

PR #297, opened 2017-01-01
Refactor tabbable.js

abouthiroppy:
Make use of `Number.isNaN` and `Array.from`.

Changes proposed:
- Replace `window.isNaN` with `Number.isNaN`
- Replace `[].slice.call` with `Array.from`

Acceptance Checklist:


abouthiroppy:
That makes sense... 

> I'm going to setup SauceLabs testing to make it known what browsers we support and document it since it's not shown anywhere currently.

Thank you 😃 

> My only concern with the transform is that it might bloat the size of the package.

I agree. OK,  I will once close this PR.

PR #299, opened 2017-01-02
Get rid of CommonJS module

abouthiroppy:
Changes proposed:
- Make use of es6 module

Acceptance Checklist:

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,2 +1,3 @@
-module.exports = require('./components/Modal');
+import Modal from './components/Modal';
 
+export default Modal;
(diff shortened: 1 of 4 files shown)


lib/index.js
```diff
@@ -1,2 +1,2 @@
-module.exports = require('./components/Modal');
-
+import Modal from './components/Modal.js';
+export default Modal;
```
abouthiroppy: I think it would be better(?) to use `export-default-from`, but this is stage-1.
What do you think?

`export-default-from`

```js
export default from './components/Modal';
```

Issue #173, opened 2016-05-01
Multiple Modals


czbaker:
When working on my current project, I discovered that it doesn't seem like `react-modal` appreciates the use of multiple `<Modal />` components in one app.

I googled around and then realized that I should use **one** `<Modal />` component and then determine through state/whatever what to display inside the modal.

I really think this should be documented, or at least mentioned in some way in the README to avoid confusion.

goldensunliu:
@czbaker you can always control when to actually render the `Modal` in its container/parent's state so that it doesn't get rendered at all

bf:
I also have the issue that modal is used deep down in some business logic, and I was a bit shocked that I ended up with 50 modal instances in my app. Is there any example how to wrap modal in it's own class or could we make it re-use the already existing HTML elements?

diasbruno:
@czbaker @bf You can make a component to a wrapped modal class. Something like this...

```javascript
class MyModal {
  render() {
    return (<Modal ...>{this.props.children}</Modal>);
  }
}
```

diasbruno:
```javascript
...
  render() {
     return (<MyModal><TheComponent /></MyModal>);
  }
...
```

or without the wrapper...

```javascript
...
  render() {
     return (<Modal><TheComponent /></Modal>);
  }
...
```

craigcosmo:
@diasbruno Hi can show how to solve this problem

I have many items on a page. I want when I click an item, a modal will show up, and that modal contains specific detail of that item

```jsx
export default class Grid extends React.Component {
	render() {
		let data = [ 
			{id:1, name: 'john', star: 4}, 
			{id:2, name: 'peter', star: 1}, 
			{id:3, name: 'sasa', star: 3} 
		]
		return (
			<div>
				{
					data.map( function(i, index){
						return (
							<div className="item">
								<span>{i.id}</span>
								<span>{i.name}</span>
								<span>{i.star}</span>
							</div>
						)
					})	
				}
				<Modal
					overlayClassName="detail-modal-overlay"
					className="detail-modal"
					isOpen={this.state.detailModal}
					onRequestClose={this.closeDetailModal.bind(this)}>
					
				</Modal>
			</div>
		)
	}
}
```

modal content should be like this when opened
```
<div className="modal-content">
	<span>{i.id}</span>
	<span>{i.name}</span>
	<span>{i.star}</span>
</div>
```

diasbruno:
@craigcosmo inside the modal add this `{props.children}`.

[...]

PR #323, opened 2017-02-15
WIP: multiple modals.

diasbruno:
This (WIP)PR is for code review for this feature and other ideas are welcome.

Changes proposed:
- This PR creates an stack of modals, so they can be on top of each other. 
- The previous modal gets the focus when the top modal is closed.
- A modal manager could be better, but for now...

Related to #173.

Acceptance Checklist:


diasbruno:
It's actually not necessary. Modals can be stacked easily, but the focus manager is not reliable to give the focus back to the previous opened modal.

PR #324, opened 2017-02-15
[change] improve reliability on focus management.

diasbruno:
This PR improves reliability to give back focus
to parent modal.

- empty

Acceptance Checklist:

### DIFF
--- a/lib/helpers/focusManager.js
+++ b/lib/helpers/focusManager.js
@@ -1,7 +1,7 @@
 import findTabbable from './tabbable';
 
+const focusLaterElements = [];
 let modalElement = null;
-let focusLaterElement = null;
 let needToFocus = false;
 
 function handleBlur () {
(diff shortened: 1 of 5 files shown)


lib/helpers/focusManager.js
```diff
@@ -30,16 +30,19 @@ function handleFocus () {
...
 export function markForFocusLater () {
-  focusLaterElement = document.activeElement;
+  focusLaterElement.push(document.activeElement);
 }
 
 export function returnFocus () {
+  let toFocus = null;
   try {
-    focusLaterElement.focus();
+    toFocus = focusLaterElement[focusLaterElement.length - 1];
```
claydiffrient: I wonder if rather than doing this... we could probably consolidate the `.pop()` to here since `pop` returns the last element of the array seems that it'd be the same thing.

Perhaps like this: 
```js
try {
  toFocus = focusLaterElement.pop();
  toFocus.focus();
  return;
}
```

lib/helpers/focusManager.js
```diff
@@ -1,7 +1,7 @@
 import findTabbable from './tabbable';
 
+const focusLaterElement = [];
```
claydiffrient: Since this is an array now... perhaps it should be changed to `focusLaterElements`?

specs/Modal.spec.js
```diff
@@ -111,6 +111,44 @@ describe('Modal', () => {
...
+    const input = (
+      <input
+        className="focus_input"
+        ref={el => (el && el.focus())}
+      />
+    );
+
+    const modal = renderModal({
+      isOpen: true,
+      onRequestClose: function cleanup () {
```
claydiffrient: Method shorthand here? 

```
onRequestClose () {

}

diasbruno:
@claydiffrient updated. Thanks!

PR #325, opened 2017-02-15
[fixed] compatibility with unstable_handleError

restrry:
in React v15 componentWillUnmount could be called without calling componentDidMount, in case when error during rendering is handled with unstable_handleError

As I can see [master version](https://github.com/facebook/react/blob/master/src/renderers/shared/stack/reconciler/ReactCompositeComponent.js#L474) should handle it carefully

Example: https://github.com/restrry/git-user/blob/react-modal/src/client.js
Playground: https://github.com/restrry/git-user/tree/react-modal

Changes are quite trivial, should I add test case?


diasbruno:
@restrry A test case would be good. :)

lib/components/Modal.js
```diff
@@ -101,6 +103,8 @@ export default class Modal extends Component {
   }
 
   componentWillUnmount () {
+    if (!this.mounted) return;
```
diasbruno: You can check if `this.node` is already filled instead of creating a flag to notify when it's mounted.

lib/components/Modal.js
```diff
@@ -86,6 +86,8 @@ export default class Modal extends Component {
     const parent = getParentElement(this.props.parentSelector);
     parent.appendChild(this.node);
     this.renderPortal(this.props);
+
+    this.mounted = true;
```
diasbruno: This line can be removed. See comment in `componentWillUnmount()`.

diasbruno:
@restrry Sorry for the delay on this. I suspect that we have issues that are related with this one.

restrry:
updated

diasbruno:
I'll apply your patch with `git format-patch -2 restrry:unstable_handleError-compat` to target `v1`.

Thank you, @restrry.

diasbruno:
Here is the commit that ended up on the tree. 28ecc0b
Released `v1.9.3`.

Issue #308, opened 2017-01-12
2+ modals on same page make an issue 


rowdyroad:
### Summary:
I what to use < body > class 'ReactModal__Body--open' to freeze background page (overflow:hidden).
But there is some issue if we have 2+ <ReactModal> on same page (component)

### Steps to reproduce:
 1. Create component
 2. Add 2+ < ReactModal > components on page
 3. Set for some of them isOpen attribute to true 

### Expected behavior:
There is class name of < body > as 'ReactModal__Body--open' if some of ReactModal component isOpen.
### Additional notes:

In 1.6.5 version we have this code:
`if (props.isOpen) {
      elementClass(document.body).add('ReactModal__Body--open');
    } else {
      elementClass(document.body).remove('ReactModal__Body--open');
    }`

So, one of < ReactModal > is open and add class to body, some is close and remove it. In Google Dev Console we can see, what react changes body tag constantly (CPU for tab is about 50% for i7).

It can be resolved(workaround) by JSX-condition wrapping and set permanently isOpen={true}

MarkMurphy:
I think one potential solution here would be to implement a `ModalManager` that keeps track of open modals. Each modal instance would have the responsibility of registering itself and it's state with the `ModalManager`.

Here's a quick and dirty example:

```js
const registry = new Set()

class Modal extends Component {
  renderPortal(props) {
    if (props.isOpen) {
      registry.add(this)
    } else {
      registry.delete(this)
    }

    if (registry.size) {
      elementClass(document.body).add('ReactModal__Body--open')
    }
    else {
      elementClass(document.body).remove('ReactModal__Body--open')
    }

    // ...
  }
}
```

diasbruno:
Marked as duplicated. #231
See PR #326.

Issue #231, opened 2016-09-23
If 2 or more modals are open and one is closed, `ReactModal__Body--open` class is erroneously removed


ameyms:
### Summary:

If two or more modals are open and one is closed, `ReactModal__Body--open` class is erroneously removed from body element
### Steps to reproduce:
1. Open a modal M1
2. Open a second modal M2
3. Close M2. The class `ReactModal__Body--open` will no longer exist on `body` element.
### Expected behavior:

The class `ReactModal__Body--open` should be active on the `body` element as long as there is at least _one_ instance of modal open
### Additional notes:

Perhaps we need to keep track of the total number of open modal instances?
https://github.com/reactjs/react-modal/blob/6550b87636546701e4aa9d90a67b4c2f431a86fe/lib/components/Modal.js#L66-L73

thesbros:
I made a (bad) decorator to fix this, if it helps anyone: https://gist.github.com/thesbros/b9ff804902d5ecc53d92caf8e81dea9b

fedeoo:
Even the [demo](https://reactcommunity.org/react-modal/) still has this bug!

iluminus:
Hi,

When do you expect this bug to be resolved?
I have the same issue and am waiting for the resolution before I upgrade to the latest version.

Thanks!

claydiffrient:
@iluminus I'm pretty sure this should be solved on the 1.7.x release.  Please give it a try and see if it works for you.  If it does, we'll close this issue.  I think it should be closed already, but I'll hold it open to get verification from you.

iluminus:
@claydiffrient I checked and it works when installing your latest version.
Thank you!

nickaversano:
I'm still experiencing this issue with multiple `ReactModal`s on the page. The `ReactModal__Body--open` gets added and then immediately removed on page reload. Inspecting the dom then shows one `ReactModalPortal` with children that `isOpen=true` for.

PR #326, opened 2017-02-15
[fixed] keep references of modals when available.

### LINKED ISSUE #218: Open two modals. Hit ESC then hit ECS but nothing happens
I am using this great idea http://stackoverflow.com/questions/35623656/how-can-i-display-a-modal-dialog-in-redux-that-performs-asynchronous-actions/35641680 in order to display modals.

In order to display them, I have this code:

``` jsx
const ModalRoot = ({ list } : ModalRootPropTypes) : any => {

  if (list.size === 0) {
    return null;
  }

  return (
    <div>
      {
        list.map(({ type, props }, index) => {
          const Component = getComponent(type);

          return (
            <Component key={index} {...props} />
          );
        })
      }
    </div>
  );
};

```

The `getComponent` is like this: 

``` jsx
const DeleteArticle = (
 <Modal>
...
</Modal>
);

const MODAL_DELETE_ARTICLE = "MODAL_DELETE_ARTICLE";

const getComponent = (type : ModalActionType) : any => {
  switch (type) {

  case MODAL_DELETE_ARTICLE:
    return DeleteArticle;
....
```

So, essentially, for 2 modals, these pieces of code generates this view:

image

The HTML code is:

image

Which is very good, correct and right. But:

Now, when I hit `ESC`, the second modal hides (which is good). Then, if I press again ESC, **nothing happens**. That's because the `Modal 1` is not logger focused. 

If I click on it and hit `ECS` it is closed.

My question is:

Can I somehow focus the next modal when one is closed? (Focus `Modal 1` when `Modal 2` is closed)

diasbruno:
keeping a reference of all modals in order to manage
when to add/remove classes and aria from elements
correctly.

solution based on @MarkMurphy's idea #308, but it uses a simple list
to work on IE (don't really support `Set`).

Fixes #218 #231 #308.

Acceptance Checklist:


ajfuller:
Hi @diasbruno, I've taken your tests from this PR and proposed an alternative solution here: https://github.com/reactjs/react-modal/pull/328

Please let me know if you have any questions or if I'm missing a case we should be testing for 😸

diasbruno:
@ajfuller awesome. I'll check it.

boonier:
@diasbruno thanks for the heads up on this. Any idea when this update might land?

ChenRoth:
can you resolve the branch conflict and submit this?

diasbruno:
@ChenRoth this needs to be ported to `v1`. I'll do it.

boonier:
@diasbruno this hasn't been released yet has it?

diasbruno:
@boonier I'll try to write the backport to `v1` today.

diasbruno:
@boonier PR is #358. Please, if you have time, review the PR.

PR #327, opened 2017-02-16
[fixed] forward onKeyDown and onClick events.

### LINKED ISSUE #184: Cannot respond to keypress events on the modal content
### Summary:

After the modal is open, focus is given to the modal content element, which is an element that is created by react-modal. When any key press is made on this element, it is not propagated to the child, so there is no way for anyone to define keypress events on the modal.
### Steps to reproduce:
1. Use demo modal, with content wrapped in parent div with onKeyPress handler.
2. Render, and try to fire keypress event.
3. Nothing happens.
### Expected behavior:

Keypress should be propagated down to the children - especially in the case that there is only one child, which is the parent of the rest of the content. Alternatively, Modal component should accept onKeyPress handler which is called if whatever keypress handler is on `.ReactModal__Content` element is not triggered (esc seems to be the only key that it's watching for). The latter seems easier to implement and more useful.
Eg left/right in image carousel.

diasbruno:
Forward onKeyDown and onClick events.

Fixes #184.

Acceptance Checklist:

### DIFF
--- a/examples/basic/app.js
+++ b/examples/basic/app.js
@@ -77,6 +77,8 @@ var App = React.createClass({
                id="test2"
                closeTimeoutMS={150}
                isOpen={this.state.modal2}
+               onKeyDown={(event) => { console.log("key pressed:", event.keyCode, event.which); }}
+               onClick={(event) => { console.log("mouse pressed:", event); }}
                onAfterOpen={() => {}}
                onRequestClose={this.closeSecondModal}>
           <p>test</p>
--- a/lib/components/Modal.js
+++ b/lib/components/Modal.js
@@ -36,6 +36,8 @@ var Modal = React.createClass({
     }),
     portalClassName: React.PropTypes.string,
     appElement: React.PropTypes.instanceOf(SafeHTMLElement),
+    onClick: React.PropTypes.func,
+    onKeyDown: React.PropTypes.func,
     onAfterOpen: React.PropTypes.func,
     onRequestClose: React.PropTypes.func,
     closeTimeoutMS: React.PropTypes.number,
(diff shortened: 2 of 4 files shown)


diasbruno:
It should not be necessary to forward this events. Just use a `<div />` wrapping the modal content.

kendallb:
Is this actually fixed? I am trying to attach a key handler and the only way I can do it is to wrap the modal in my own div and attach the handler to that, but that is not a clean solution. I tried adding an onKeyDown handler to the modal itself, but that did not do anything? Suggestions?

diasbruno:
Hi @kendallb, the idea is to separate the keyboard/mouse events from the modal and the component that goes in the modal.

kendallb:
Well that was my original approach as well but that does not work in all situations because it is outside the modal block so it stays in the DOM where you use the component rather than moving to the end of the DOM like the modal block does.

It would be a lot better if it would forward events onto my event handler for me.

diasbruno:
It's a trade off to keep the separation between components. Of course, we can review this approach and implement the forward method. 
You can have a look on issue #184 for further information.

kendallb:
That one is closed also and I did implement the solution fritz mentioned and it works. It’s just not a logical solution so if someone else runs across the problem without finding the ticket they won’t figure it out. My initial assumption was I could just register key down with the modal and it would work, and it did not. If that worked, most people would figure that out quickly. Probably the same goes for mouse events.

diasbruno:
Good point. I'll have a look on some options to either merge this PR or update the documentation explaining why react-modal don't allow to hook on input events.

diasbruno:
Thanks for the ideas.

kendallb:
Thanks for the module! Apart from these minor issues it works great. Took me a bit to pull the styles I used to use for jQuery UI to work with my react dialogs, but once I got it working it works really well.

PR #328, opened 2017-02-17
[fixed] Multiple modals on page causing contradictory class updates

### LINKED ISSUE #218: Open two modals. Hit ESC then hit ECS but nothing happens

ajfuller:
Changes proposed:
- Modals should only update if they are either opening or closing. If they will remained closed then there is no reason to run the `renderPortal()` method.

Fixes #218 #231 #308

Used the tests from https://github.com/reactjs/react-modal/pull/326

Acceptance Checklist:


specs/Modal.spec.js
```diff
@@ -234,25 +244,42 @@ describe('Modal', () => {
...
+  });
+
+  it('should remove class from body when no modals opened', () => {
+    const expectedBodyClass = 'ReactModal__Body--open';
+    const classList = document.body.classList;
+    renderModal({ isOpen: true });
+    renderModal({ isOpen: true });
+    expect(classList.contains(expectedBodyClass)).toBe(true);
+    unmountModal();
+    expect(classList.contains(expectedBodyClass)).toBe(false);
```
diasbruno: Here is the problem. 2 modals were opened, one of those is unmounted, so `document.body` should actually have the class `ReactModal__Body--open`.

The class should be removed only if all modals were closed.

ajfuller: Ahh you're right I reversed that test inadvertently 😿 

So the design decision is to allow for modal stacking then? I naively assumed from both an accessibility and UX perspective it should only allow one modal at a time to show. Would there be any consideration to disallowing stacking or is that firmly part of the component?

Thanks for the review and help understanding!

diasbruno: I don't see any accessibility and UX problem on stacking modal (regarding that we get to manage the focus and open/exit modals correctly).

Modal should block the UI for a single context or scope (my view). Different from a `window` or `popup`.

diasbruno: I think the correct fix for this issue is to have some sorta manager to deal with this.

ajfuller: We have pretty strict guidelines around modals not being stackable, but I understand that it is not a universal guideline, so keeping the component flexible makes sense.

I'll likely continue to use the `shouldComponentUpdate` approach within my composition of the component, but feel free to close if it's not the direction you're looking for 👍

diasbruno: If using redux or a normal component state, you can set an state to avoid opening more modals at the same time. 

Just to illustrate the idea:

```javascript
{ currentModal: null } // can open
```

```javascript
{ currentModal: 'my-modal-1' } // already in use
```

diasbruno:
Related issue #335.

adjohnson916:
Relates to #358.

diasbruno:
@ajfuller Thank you so much for the attention to this.

PR #330, opened 2017-02-20
[changed] use object className and overlayClassName prop to override …

cassln:
Changes proposed:
- use object className and overlayClassName prop to override the default content and overlay classes;
- use bodyOpenClassName to override body class name when modal opened.

For overriding default content and overlay classes you can pass object
with three required properties: `base`, `afterOpen`, `beforeClose`.
```jsx
<Modal
  ...
  className={{
    base: 'myClass',
    afterOpen: 'myClass_after-open',
    beforeClose: 'myClass_before-close'
  }}
  overlayClassName={{
    base: 'myOverlayClass',
    afterOpen: 'myOverlayClass_after-open',
    beforeClose: 'myOverlayClass_before-close'
  }}
  ...
>
```
Old behavior is saved, when passed a string. 

Acceptance Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,26 @@ you can pass `className` and `overlayClassName` props to the Modal.  If you do
 this then none of the default styles will apply and you will have full control
 over styling via CSS.
 
+If you want to override default content and overlay classes you can pass object
+with three required properties: `base`, `afterOpen`, `beforeClose`.
+
+```jsx
+<Modal
+  ...
+  className={{
+    base: 'myClass',
+    afterOpen: 'myClass_after-open',
+    beforeClose: 'myClass_before-close'
+  }}
+  overlayClassName={{
+    base: 'myOverlayClass',
+    afterOpen: 'myOverlayClass_after-open',
+    beforeClose: 'myOverlayClass_before-close'
+  }}
+  ...
+>
+```
+
 You can also pass a `portalClassName` to change the wrapper's class (*ReactModalPortal*).
 This doesn't affect styling as no styles are applied to this element by default.
 
@@ -112,6 +132,7 @@ function getParent() {
 ### Body class
 When the modal is opened a `ReactModal__Body--open` class is added to the `body` tag.
 You can use this to remove scrolling on the the body while the modal is open.
+You can also pass a `bodyOpenClassName` to change the default class.
 
 ```CSS
 /* Remove scroll on the body when react-modal is open */
--- a/specs/Modal.spec.js
+++ b/specs/Modal.spec.js
@@ -175,6 +175,32 @@ describe('Modal', () => {
     expect(modal.portal.overlay.className.indexOf('myOverlayClass')).toNotEqual(-1);
   });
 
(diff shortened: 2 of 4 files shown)


diasbruno:
hey @cassln, this PR looks good, but why would you want to replace the class names?
what if we could have something like this:

sass or less style (replace `react-modal-*` with the actual classes).

```sass
.{my-class}.react-modal-base { ... }
.{my-class}.react-modal {
  .react-modal-overlay { ... }
  .react-modal-content { ... }
}
```

cassln:
I want to replace the class names because of I'm sticking [BEM](https://en.bem.info/) notation. And cascades like `.{my-class}.react-modal-base { ... }` are working, but it is wrong for BEM.

Also, this feature makes possible getting clean markup for react-modal in result. :)

cassln:
What about merge the feature? :) I really need this in working project.

diasbruno:
@cassln don't wait, use it from your fork. :)

diasbruno:
Be patient. I do think this PR is good. there is some working going on to get a version in between `1.6.5` and `2.0`.

lib/components/ModalPortal.js
```diff
@@ -30,8 +22,22 @@ export default class ModalPortal extends Component {
...
+        beforeClose: PropTypes.string.isRequired
+      })
+    ]),
+    overlayClassName: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.shape({
+        base: PropTypes.string.isRequired,
+        afterOpen: PropTypes.string.isRequired,
+        beforeClose: PropTypes.string.isRequired
+      })
```
diasbruno: It seems that only the `base` of the classes change. Maybe we can short this changes by keeping the suffixes unchanged like `{ReactModal}_before-open` and `{ReactModal}_after-open`. Any thoughts?

cassln: The suffixes are not flexible. There are BEM [alternative naming schemes](https://en.bem.info/methodology/naming-convention/#alternative-naming-schemes). The suffixes make impossible of using them. Also, we can pass props like this (without BEM example):

```javascript
{
  base: 'modal',
  afterOpen: 'open',
  beforeClose: 'close'
}
```

cassln:
> there is some work going on to get a version in between 1.6.5 and 2.0

Do you have ability to release version between 1.6.5 and 2.0. This It would be very helpful. :+1:

claydiffrient:
@cassln 1.7.0, 1.7.1, and 1.7.2 have been released.

cassln:
The branch has been rebased.

viniciusffj:
@cassln @diasbruno Is there a chance to this get merged? I really need the `bodyOpenClassName` funcionality

diasbruno:
cc @claydiffrient @cassln

pamelipluas:
@cassln @diasbruno Hi everyone! Any update about this pl? In my project we are using this lib and we really need can change the body classname

abdennour:
Why this PI is not merged yet ?

PR #348, opened 2017-03-06
Remove .bind(this) from removePortal call

### LINKED ISSUE #344: getting bind warning from react after 1.7.0
### Summary:

I'm getting this warning from react in the console:

`Warning: bind(): You are binding a component method to the component. React does this for you automatically in a high-performance way, so you can safely remove this call. See Modal`

Here is the call where's it happening (look for the arrow in comments on right side)

```
  componentWillUnmount: function() {
    if (this.props.ariaHideApp) {
      ariaAppHider.show(this.props.appElement);
    }

    var state = this.portal.state;
    var now = Date.now();
    var closesAt = state.isOpen && this.props.closeTimeoutMS
      && (state.closesAt
        || now + this.props.closeTimeoutMS);

    if (closesAt) {
      if (!state.beforeClose) {
        this.portal.closeWithTimeout();
      }

      setTimeout(this.removePortal.bind(this), closesAt - now);   /// <-- This line
    } else {
      this.removePortal();
    }
  },
```

### Steps to reproduce:

I think it happens when I unmount a modal when it's still disappearing.

mreishus:
This fixes this warning from react in the console without breaking functionality:

```
Warning: bind(): You are binding a component method to the component. React does this for you automatically in a high-performance way, so you can safely remove this call. See Modal
```

Fixes #344.

This doesn't fix a bug or change the API, it should be basically a "silent" change.

Acceptance Checklist:

### DIFF
--- a/lib/components/Modal.js
+++ b/lib/components/Modal.js
@@ -94,7 +94,7 @@ var Modal = React.createClass({
         this.portal.closeWithTimeout();
       }
 
-      setTimeout(this.removePortal.bind(this), closesAt - now);
+      setTimeout(function() { this.removePortal(); }, closesAt - now);
     } else {
       this.removePortal();
     }


diasbruno:
This test has failed:
`keeps the modal in the DOM until closeTimeoutMS elapses.`

Use `npm run test`.

mreishus:
My fix was actually bad.  It was causing `removePortal` to be run at the wrong time.  I added a closure around it (amended commit and forced push to keep it all in one commit).

mreishus:
erm, I think I shouldn't have used es6.

claydiffrient:
Chances are you'll probably want to just rebase on v1 and do a forced push to clean things up though.

mreishus:
OK, redid off `v1` branch, fixed tests and lint.  I did use the `() =>` es6 construction tho.

mreishus:
Don't merge this please, it causes the same uglifyJS error that happened in issue 336.

claydiffrient:
I was thinking that it might @mreishus, my guess is it's the `() =>`

mreishus:
done

claydiffrient:
Released in 1.7.2

lib/components/Modal.js
```diff
@@ -94,7 +94,7 @@ var Modal = React.createClass({
         this.portal.closeWithTimeout();
       }
 
-      setTimeout(this.removePortal.bind(this), closesAt - now);
+      setTimeout(function() { this.removePortal(); }, closesAt - now);
```
adjohnson916: I think `removePortal` won't be available on `this` in `setTimeout` which will have `this` context of `window`? If you don't use arrow function, it won't scope capture.

diasbruno: yes, you are right!

Issue #351, opened 2017-03-09
Can not close Modal when click on overlay (Please update the note in readme)


SLyHuy:
### Summary: Can not close Modal when click on overlay
└─┬ react-modal@1.7.2 
  ├── element-class@0.2.2 
  ├── exenv@1.2.0 
  └── lodash.assign@4.2.0 

### Steps to reproduce:

 1. Remove **shouldCloseOnOverlayClick** prop --> Click on overlay --> nothing happen.
 2. Add shouldCloseOnOverlayClick={true} --> Click on overlay --> nothing happen.
 3. Add shouldCloseOnOverlayClick={false} --> Click on overlay --> nothing happen.

### Expected behavior: 

 1. Remove **shouldCloseOnOverlayClick** prop --> Click on overlay --> close modal.
 2. Add shouldCloseOnOverlayClick={true} --> Click on overlay --> close modal.
 3. Add shouldCloseOnOverlayClick={false} --> Click on overlay --> nothing happen.

SLyHuy:
I read source code, and found:

![screen shot 2017-03-09 at 16 07 58](https://cloud.githubusercontent.com/assets/1643397/23743331/d01d17da-04e2-11e7-95c4-b5483de11d92.png)

Then it checks `ownerHandlesClose` and it needs `onRequestClose` prop. Then I have to pass manual function to change state for open or close modal.

Please put the note to Readme please.

diasbruno:
@SLyHuy Can you add this note to the readme?
It's required to pass a function to control the state of the modal.

SLyHuy:
@diasbruno Here is pull request https://github.com/reactjs/react-modal/pull/352

claydiffrient:
#352 closed this issue.

PR #352, opened 2017-03-10
Update README for issue close modal when user click on overlay

SLyHuy:
https://github.com/reactjs/react-modal/issues/351


README.md
```diff
@@ -220,7 +220,7 @@ When using React Test Utils with this library, here are some things to keep in m
 - You need to use the `.portal` property, as in `ReactDOM.findDOMNode(renderedModal.portal)` or `TestUtils.scryRenderedDOMComponentsWithClass(Modal.portal, 'my-modal-class')` to acquire a handle to the inner contents of your modal.
 
 By default the modal is closed when clicking outside of it (the overlay area). If you want to prevent this behavior you can
```
claydiffrient: Hmm... maybe this whole section should read more like:

```
By default the modal will attempt to close when clicking outside of it (on the overlay area).  Because the open and closed state of the modal should be controlled outside of the modal, you need to provide the `onRequestClose` function prop to actually set the state.  If you want to prevent the overlay close behavior, you should pass the `shouldCloseOnOverlayClick` props as `false`.

SLyHuy: Ok.

PR #354, opened 2017-03-11
focus on content after non close

halt-hammerzeit:
My modal has a `busy` state and when it's `busy` it cannot be allowed to close.
So if in `onRequestClose` `busy` is `false` it plays an animation and doesn't close and returns `false`.
The issue is that the focus is lost.
This commit fixes the lost focus bug.


halt-hammerzeit:
I found another way: `findDomNode(this.ref).parentNode.focus()`

Issue #359, opened 2017-03-22
aria-hidden=true on body breaks Accessibility <body class="ReactModal__Body--open" aria-hidden="true">


pauljadam:
### Summary:
Screen reader users cannot interact with any modal content because aria-hidden=true is incorrectly placed on the body tag.

### Steps to reproduce:

 1. Open a test modal, https://reactcommunity.org/react-modal/examples/minimal.html.
 2. Activate modal dialog.
 3. Check that aria-hidden=true is incorrectly placed on the body tag of the modal's html document.

### Expected behavior:

Screen reader user is trapped inside the modal dialog.

### Actual behavior:

Screen reader user can't interact with anything on the page at all because of aria-hidden=true on the body.

pauljadam:
There's a new aria-modal=true attribute that should be placed on the modal dialog which could remove the need for aria-hidden.

diasbruno:
@pauljadam can you do a little research on this topic and/or provide more sources for this?

pauljadam:
http://w3c.github.io/aria/aria/aria.html#aria-hidden

aria-hidden (state)

Indicates whether the element is exposed to an accessibility API.

If the body is hidden, <body aria-hidden=true> then the whole body tag is no longer exposed to an Accessibility API. You only want to hide the main content container(plus header/footer) of the page not the entire <body> tag.

pauljadam:
http://w3c.github.io/aria-practices/#dialog_modal

This demo uses aria-modal=true rather than aria-hidden. http://w3c.github.io/aria-practices/examples/dialog-modal/dialog.html

marcysutton:
I can confirm this behavior. It also looks like focus management isn't happening on the modal, and it's missing the dialog role (though those could be separate issues). Here's how to fix it:

1. When the modal opens, put `aria-hidden="true"` on **sibling element(s)** to the dialog instead of the body, and change it back to false when it closes.
2. Send focus to the dialog close button on open.
3. Send focus back to the triggering button on close.

Following the ARIA practices guide is a great suggestion!

diasbruno:
Thank you so much @pauljadam and @marcysutton.

claydiffrient:
[...]

Issue #133, opened 2016-03-02
Error: You must set an element with `Modal.setAppElement(el)` to make this accessible


vonwao:
I get error, but CHANGELOG says:

v0.5.0 - Tue, 22 Sep 2015 19:19:44 GMT
[changed] setAppElement method is now optional.

So I'm confused.

Error: react-modal: You must set an element with `Modal.setAppElement(el)` to make this accessible
    at validateElement (index.js:224)
    at show (index.js:211)
    at Object.toggle (index.js:219)

zires:
Same problem here

yachaka:
Hello,

Same problem here.  
The whole problem is the scripts being loaded at the top of the page. `react-modal` was trying to set the app element to `document.body`, while it doesn't exist yet. If you did the same, within your code, it resulted as the same as your script was probably loaded before the DOM too.

I solved it by calling it in the lifecycle method `componentWillMount`:
```js
componentWillMount() {
    Modal.setAppElement('body');
 },
```

I can create a pull request solving this issue if you wish to?

ryanflorence:
woah, what? that should not be optional, that's the whole point of this library.

stkrzysiak:
Thank you @yachaka for that snippet, solved my issue.  I ran into this when running tests, fwiw.

yachaka:
@ryanflorence The doc is precising that it's optional, and that by default react-modal is anchored to `document.body` (which isn't, as the DOM isn't existing yet):
```
/*
By default the modal is anchored to document.body. All of the following overrides are available.

* element
Modal.setAppElement(appElement);

* query selector - uses the first element found if you pass in a class.
Modal.setAppElement('#your-app-element');

*/
```
I guess the doc/or the code has to change?

ungoldman:
I only got this error when we switched to using webpack-dev-server. Very weird.

ungoldman:
Update: injecting the bundled application JS into `<body>` instead of `<head>` seems to have solved it for me for now, but I think ideally react-modal should wait until the DOM is loaded to try attaching to `document.body` (if that's what's going on).

RobinClowers:
The `appElement` is only used in one place: https://github.com/reactjs/react-modal/blob/master/lib/components/Modal.js#L75, and isn't documented at all. I think that line should be using the closed over `appElement` variable instead.

radixhound:
[...]

PR #360, opened 2017-03-24
[changed] Replace appElement with getAppElement

### LINKED ISSUE #287: Remove Modal.setAppElement?
I'm thinking perhaps we should get rid of `Modal.setAppElement`.  It's slightly odd and trips many people up when trying to use the library.  There's also been the odd `appElement` prop that has been around.

I think we should remove the `appElement` prop replacing it with a `getAppElement` function prop similar to what was done for the `parentElement` stuff in 1.6.x.  I'm almost certain this would help with things like #174 is having issues with.

Thoughts?

claydiffrient:
This makes getAppElement a required prop as well as makes
it a function that will be called expecting a DOMElement.

closes #287

This also takes some inspiration from #359 for handling arrays
of objects.

Fixes #[issue number].

Changes proposed:
- Remove appElement in all it's forms
- Add getAppElement as a function prop

  - If you had specified an appElement via `Modal.setAppElement`,
    then you need to convert that to a getAppElement prop on the
    modal, this should be a function that returns either a single
    element or an array of elements.
  - If you had nothing specified you will need to add the getAppElement
    element to prevent breakages.

Acceptance Checklist:

### DIFF
--- a/src/helpers/ariaAppHider.js
+++ b/src/helpers/ariaAppHider.js
@@ -1,37 +1,24 @@
-let globalElement = typeof document !== 'undefined' ? document.body : null;
-
 function validateElement (appElement) {
-  if (!appElement && !globalElement) {
-    throw new Error('react-modal: You must set an element with `Modal.setAppElement(el)` to make this accessible');
+  if (!appElement) {
+    throw new Error('react-modal: Setting an getAppElement function is required');
   }
 }
 
-export function setElement (element) {
-  let newElement = element;
-  if (typeof newElement === 'string') {
-    const el = document.querySelectorAll(element);
-    newElement = 'length' in el ? el[0] : el;
+export function toggle (appElement, value) {
+  validateElement(appElement);
+  if (Array.isArray(appElement)) {
+    appElement.forEach((ae) => {
+      ae.setAttribute('aria-hidden', value);
+    });
+  } else {
+    appElement.setAttribute('aria-hidden', value);
   }
-  globalElement = newElement || globalElement;
-  return globalElement;
 }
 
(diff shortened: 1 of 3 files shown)


src/helpers/ariaAppHider.js
```diff
@@ -1,37 +1,35 @@
...
   validateElement(appElement);
-  (appElement || globalElement).removeAttribute('aria-hidden');
+  if (Array.isArray(appElement)) {
+    appElement.forEach((ae) => {
+      ae.removeAttribute('aria-hidden');
+    });
+  } else {
+    appElement.removeAttribute('aria-hidden');
+  }
 }
```
diasbruno: `show()` has the same effect if it sets `ae.setAttribute('aria-hidden', 'false');`, so `hide()` and `show()` could be`toggle(appElement : DOM, value : bool)`.

claydiffrient: Good call :)

dminuoso:
This solution elegantly tackles #133 too.

dminuoso:
One thing that I've been thinking about is the change to always require a getAppElement.

Does defaulting to `() => document.body` make sense perhaps?

claydiffrient:
@dminuoso No unfortunately it doesn't make sense.  If you do that then all the body ends up hidden from screen readers which definitely makes it non-accessible.  We've got that in v1 currently and there are lots of sites out there that get completely hidden from screen readers because of it.

priyajeet:
@claydiffrient are there any plans to back port this to version 1 for issue #133. Or when will V2 have a release (beta or otherwise). Thanks!

PR #364, opened 2017-03-31
[added] bool shouldCloseOnEsc to prevent modal closing on 'esc' press.

creatyvtype:
Fixes #[363].

Changes proposed:
- Add boolean shouldCloseOnEsc
- Add test for said bool

N/A

Acceptance Checklist:
- [X] If the commit message has [changed] or [removed], there is an upgrade path above.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -61,6 +61,10 @@ import ReactModal from 'react-modal';
   */
   shouldCloseOnOverlayClick={true}
   /*
+    Boolean indicating if the 'esc' key should close the modal
+   */
+  shouldCloseOnEsc={true}
+  /*
     String indicating the role of the modal, allowing the 'dialog' role to be applied if desired.
   */
   role="dialog"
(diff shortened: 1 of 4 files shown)


claydiffrient:
@creatyvtype Thanks for the contribution.  Unfortunately, this isn't a contribution we'll accept at this moment.  Having the escape key close the modal is fundamental to making the modal accessible.  It's the expected behavior for modals according to https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal.  In order to encourage accessibility and standardization with the best user experience possible, we believe that keeping all modals closing with the escape key would be best.

creatyvtype:
I ended up creating my own anyway. But consider this: if the modal can't close on outside click, the intended behavior could be that there is a required action inside the modal to continue. If they can get around it with pressing 'esc' you no longer have that option in your layout. 

Majd

PR #381, opened 2017-04-20
Support SSR by checking typeof window.

vernondegoede:
Fixes #[issue number].

Changes proposed:
- Check if typeof `window` is undefined. If that's true -> don't rely on global `document` variable.

- Core API shouldn't be affected with this code change.

Acceptance Checklist:

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -79,6 +79,10 @@ export default class Modal extends Component {
   }
 
   componentDidMount () {
+    if (typeof window === 'undefined') {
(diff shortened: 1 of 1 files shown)


claydiffrient:
This shouldn't be necessary to support SSR.  `componentDidMount` is only called on the client side where `window` is always defined.

vernondegoede:
You're right. Somehow my tests have `"testEnvironment": "node"` but the lifecycle hook `componentDidMount` was still called.

PR #395, opened 2017-06-07
[change] deprecate getParentElement in favor of findDOMNode.

diasbruno:
`findDOMNode` is available since React v0.14 and it's the minimum
version for `react-modal`.

To keep things simple, each Modal can be placed on any level
of the react's tree and it will do a proper clean up when it's
time to unmount the component. Also, it make things simple enough
to deal with overlapped modal.

Changes proposed:
- Use a `div` provided by the instance of <Modal /> to setup things.
- Each `<Modal />` can perform its own clean up.

- Remove `getParentElement`.

Acceptance Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,24 +107,8 @@ The default styles above are available on `Modal.defaultStyles`. Changes to this
 object will apply to all instances of the modal.
 
 ### Appended to custom node
-You can choose an element for the modal to be appended to, rather than using
-body tag. To do this, provide a function to `parentSelector` prop that return
-the element to be used.
-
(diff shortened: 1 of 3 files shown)


diasbruno:
needs some testing with different `react`and `react-dom` versions.

diasbruno:
I believe It's only needed to test with `react@0.14`.
Some version of `react` has a bug on `findDOMNode`, but it should be better to ask to upgrade for a new version.

diasbruno:
This patch need rework.

PR #410, opened 2017-06-14
Remove deprecation warning in react 15.6.0 about React.DOM.noscript

ingro:
As the title, this change remove the soon-will-be-deprecated use of React.DOM factories with the new drop-in package `react-dom-factories` instead.

I also added `cross-env` package to be able to run tests on windows.

Nothing needed.

Acceptance Checklist:


diasbruno:
Hi @ingro, can you have a look at PR #395? Hope the could use it to simplify a little bit (the portal mechanism).

ingro:
Hello @diasbruno , I wasn't aware of that PR, your solution is surely better on the long-term, I wanted just to get rid of a deprecation notice! :P

diasbruno:
Yeah, that's ok. :) 
Do you mind help me with some tests? I think it just need to be tests on `react@014`.

package.json
```diff
@@ -14,7 +14,7 @@
   },
   "scripts": {
     "start": "./node_modules/.bin/webpack-dev-server --inline --host 127.0.0.1 --content-base examples/",
-    "test": "NODE_ENV=test karma start"
+    "test": "cross-env NODE_ENV=test karma start"
```
diasbruno: This is out of scope.

andrewjmead:
While the warning about React.DOM.noscript is gone. There is still a warning related to React.DOM.div. Both are related to the new react-dom-factories depreciation. It's coming from [this](https://github.com/reactjs/react-modal/blob/v1.9.4/lib/components/ModalPortal.js#L2).

You should be able to fix it by modifying lib/components/ModalPortal.js:

```
// Require lib
var DOMFactories = require('react-dom-factories');

// Then replace this
var div = React.DOM.div;
// With this
var div = DOMFactories.div;
```

### New warning

```
Warning: Accessing factories like React.DOM.noscript has been deprecated and will be removed in v16.0+. Use the react-dom-factories package instead.  Version 1.0 provides a drop-in replacement.
```

### Previous Warning

```
Warning: Accessing factories like React.DOM.div has been deprecated and will be removed in v16.0+. Use the react-dom-factories package instead.  Version 1.0 provides a drop-in replacement. For more info, see https://fb.me/react-dom-factories
```

diasbruno:
Released `v1.9.6`.
@ingro I've added @andrewjmead changes to your PR and removed `cross-env`.
Your patch is 91e1a67.
Thank you.

ingro:
Thank you! Sorry but I couldn't reply earlier!

diasbruno:
No problem, @ingro. We are close to get `v2.0.0` merged, so I'm trying to get the last PRs merged, because I always have to rebase #412. :)
Please, open a PR for the `cross-env`.

ingro:
Done! :D

PR #414, opened 2017-06-15
Added cross-env to run tests on windows

ingro:
Run the test script through `cross-env` to be able to run them on windows.


diasbruno:
Released `v1.9.7`.

PR #417, opened 2017-06-16
[added] Build step for compiling to CommonJS modules

ajfuller:
Ran into an issue today while trying to update to version 2.0 with errors throwing because of the the static property on the class. Would like to avoid running babel against anything in node_modules, so ideally we'd have a transpiled `main` field and a `module` field that points to an ES module build.

The following changes are based on the build steps from Redux repo: https://github.com/reactjs/redux/blob/master/package.json

Changes proposed:
- Move all main source code under `src` so that `lib` and `es` can be used for the built files
- Entry `index.js` is now an ES export
- ~~Move `create-react-class` to dev dependency as it appears to be no longer necessary in the main library~~ Now removed entirely

I've tested this locally with `npm link` and everything seems to be working as expected.

Note that there are more build changes that we could make as well, but this solves the initial issue I was seeing 😸 

Acceptance Checklist:


.babelrc
```diff
@@ -1,3 +1,21 @@
 {
-  "presets": ["es2015", "stage-2", "react"]
+  "presets": ["react"],
+  "env": {
+    "commonjs": {
+      "plugins": [
```
ajfuller: Supposedly the env and non-env configs should merge, but I was seeing the below issue when I tried to consolidate. Any help is appreciated.

Tried this:
```js
{
  "presets": ["react"],
  "plugins": [
      "transform-class-properties", 
      "transform-object-rest-spread"
  ],
  "env": {
    "commonjs": {
      "presets": ["env"]
    },
    "es": {
      "presets": [
        ["env", { "modules": false }]
      ]
    }
  }
}
```

Got this:
image

ajfuller:
Will wait to squash commits until I address any review comments

diasbruno:
Awesome, @ajfuller. This seems close to commit I forgot to port from the old `master` (9cb8441).
You can use it for reference.

.babelrc
diasbruno: I think we can use the old `.babelrc` to fix `v2.0.0`. I think we need to use `latest` instead of `stage-2`.

diasbruno: `latest` seems to be not enough for this.

diasbruno:
`yarn.lock` can be removed as the build pipeline will add it when making the release commit. I'll update the documentation explaining the build and publish pipeline.

.babelrc
diasbruno: Actually after some tests, `.babelrc` is working as expected, but the `main` on package.json is pointing to the non-compiled version. You change to compile from `src` to `lib` should fix without to much change (I believe).

package.json
```diff
@@ -55,14 +56,13 @@
     "react-addons-test-utils": "^15.0.0",
     "react-dom": "^15.0.0",
     "rf-release": "0.4.0",
-    "rimraf": "^2.5.4",
+    "rimraf": "^2.6.1",
```
diasbruno: `rimraf` dependency can be removed.

.gitignore
```diff
@@ -8,3 +8,7 @@ _book
 *.diff
 examples/__build__
 coverage
+
+## Built folders
+es
+lib
```
diasbruno: I think `lib` can't be ignored. You can remove `es`.

ajfuller:
@diasbruno I can remove, but wouldn't you want the yarn.lock to be up-to-date for development as well? I'll wait for the documentation updates 😄

diasbruno:
The new pipeline will guarantee that what has been changed on `package.json` will get on the new release.

ajfuller:
@diasbruno I've merged [your PR](https://github.com/ajfuller/react-modal/pull/1) with mine and addressed the outstanding comments 👍

ajfuller:
@diasbruno everything is passing CI now 👍

diasbruno:
I think we are ready to ship! Thank you so much @ajfuller.

diasbruno:
Released `v2.0.1`.

PR #431, opened 2017-06-19
[fixed] Eliminate unnecessary renders when modal is closed

ajfuller:
Fixes #[issue number].

Changes proposed:
- Early return if the modal is closed and will remain closed. This prevents us from doing additional DOM check and running `renderPortal` again unnecessarily. This was my original intention with https://github.com/reactjs/react-modal/pull/328, but is greatly simplified with v2 code.

I wasn't exactly sure if we wanted another test as all current tests are passing and this one (https://github.com/reactjs/react-modal/blob/master/specs/Modal.spec.js#L233) covers the scenario of multiple modals rendering. 

Acceptance Checklist:

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -108,8 +108,12 @@ export default class Modal extends Component {
   }
 
   componentWillReceiveProps(newProps) {
-    if (newProps.isOpen) refCount.add(this);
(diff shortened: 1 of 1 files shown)


diasbruno:
Released `v2.0.6`.

PR #436, opened 2017-06-23
[fixed] Class names being added to body during render of non-opened modals

ajfuller:
#### Modal reference tracking
There is an issue where class names are being added to document.body during render of non-opened modals. With only one type of `bodyOpenClassName` it was "working", but if you added a custom `bodyOpenClassName` it was causing that value to be added regardless of `isOpen` status.

See the tests to replicate the behavior against current codebase.

This PR creates a new data structure for the modal references. Instead of an array of modals, we need to know which open modals are associated with each `bodyOpenClassName`.

Proposed structure is:
```js
modals = {
    "bodyClassName1": [ modal1Ref, modal3Ref ],
    "bodyClassName2": [ modal2Ref ],
    "bodyClassName3": [ modal4Ref ],
}
```

- Additionally fixed an issue where `aria-hidden` was being removed from the `appElement` before all open modals were actually closed.
- Also refactored some of the demo app so I could manually see this functionality working

I'm not a huge fan of the side-effects when manipulating the `modals` object, but it's consistent with the current code, and it works 💅 

Acceptance Checklist:


diasbruno:
@ajfuller This PR does a lot of stuff, maybe we can separate the fix from this new structure. It's better to integrate small PRs.

diasbruno:
In order to avoid adding too much complexity on `react-modal`, I think we need to add some rules to use this property since it can be really problematic.

ajfuller:
@diasbruno I can remove the `examples/basic/app.js` changes for now. I left them in because the changes helped validate the broken scenarios were actually working now.

Are there any specific changes you're concerned about? The majority of changes are consolidating the instance tracking and associated tests. I'm more than happy to change the implementation, but it was the only way I could find to make it pass the tests.

diasbruno:
We can simplify the structure by changing the value to be a `counter`. This way we can just increment and decrement the counter for each class name.

```js
<Modal isOpen={true} bodyOpenClassName={A} />
<Modal isOpen={false} bodyOpenClassName={A} />
<Modal isOpen={false} bodyOpenClassName={B} />

{ "A": 1, "B": 0 }
```

The `refCount` api can be updated to something like:

```js
// register/increment
add(bodyOpenClassName);

// unregister/decrement
remove(bodyOpenClassName);

// check if a class name must be in `document.body`
active(bodyOpenClassName) -> Boolean 
```

It seems better to handle toggle the class name and aria in the `Modal` class, otherwise it'll be hidden inside the `refCount`. Probably we need to improve both methods `renderPortal` and `removePortal`.

We also could use `componentWillReceiveProps` to throw an exception if `bodyOpenClassName` changes. Perhaps, it would be better to treat this as a violation.

It seems the way to go since we can't no longer rely on the reference of the instance neither the value of `bodyOpenClassName` (if it can change).

ajfuller:
@diasbruno good call, on the counter. To make that work I've changed the code to be in `ModalPortal` component instead. That way we're only running `refCount.add()` when the modal is actually opening and `refCount.remove()` when the modal is closing, not just whenever `renderPortal` is being called.

diasbruno:
This is really nice. Letting `Modal` be just a proxy class.

specs/Modal.spec.js
```diff
@@ -166,6 +166,7 @@ describe('State', () => {
     expect(
       moverlay(modal).className.includes('myOverlayClass')
     ).toBeTruthy();
+    unmountModal();
```
diasbruno: `unmountModal` shouldn't be necessary. `afterEach` should deal with clean up. I'll have a look on this.

specs/Modal.spec.js
```diff
@@ -233,11 +235,31 @@ describe('State', () => {
...
     unmountModal();
-    expect(!isBodyWithReactModalOpenClass()).toBeTruthy();
+    expect(isBodyWithReactModalOpenClass()).toBeFalsy();
+  });
+  
+  it('should not add classes to document.body for unopened modals', () => {
+    renderModal({ isOpen: true });
+    expect(isBodyWithReactModalOpenClass()).toBeTruthy();
+    renderModal({ isOpen: false, bodyOpenClassName: 'testBodyClass' });
+    expect(isBodyWithReactModalOpenClass('testBodyClass')).toBeFalsy();
```
diasbruno: Expectation should be always `True`. 

`expect(!isBodyWithReactModalOpenClass('testBodyClass')).toBeTruthy();`

specs/Modal.spec.js
```diff
@@ -233,11 +235,31 @@ describe('State', () => {
...
+  });
+
+  it('should not remove classes from document.body when rendering unopened modal', () => {
+    renderModal({ isOpen: true });
+    expect(isBodyWithReactModalOpenClass()).toBeTruthy();
+    renderModal({ isOpen: false, bodyOpenClassName: 'testBodyClass' });
+    renderModal({ isOpen: false });
+    expect(isBodyWithReactModalOpenClass('testBodyClass')).toBeFalsy();
```
diasbruno: `expect(!isBodyWithReactModalOpenClass('testBodyClass')).toBeTruthy();`

specs/Modal.spec.js
```diff
@@ -233,11 +235,31 @@ describe('State', () => {
   it('remove class from document.body when no modals opened', () => {
     renderModal({ isOpen: true });
     renderModal({ isOpen: true });
+    
     expect(isBodyWithReactModalOpenClass()).toBeTruthy();
     unmountModal();
     expect(isBodyWithReactModalOpenClass()).toBeTruthy();
     unmountModal();
-    expect(!isBodyWithReactModalOpenClass()).toBeTruthy();
+    expect(isBodyWithReactModalOpenClass()).toBeFalsy();
```
diasbruno: `expect(!isBodyWithReactModalOpenClass()).toBeTruthy();`

src/helpers/refCount.js
```diff
@@ -1,19 +1,25 @@
...
 
-export function count() {
-  return modals.length;
+export function count(bodyClass) {
+  return modals[bodyClass];
+}
+
+export function totalCount() {
+  const count = Object.keys(modals)
+    .reduce((acc, curr) => acc + modals[curr], 0);
```
diasbruno: This can be inlined with `return`.

src/components/ModalPortal.js
```diff
@@ -62,16 +69,27 @@ export default class ModalPortal extends Component {
...
+        console.warn(
+          'React-Modal: "bodyOpenClassName" prop has been modified. ' + 
+          'This may cause unexpected behavior when multiple modals are open.'
+        );
+      }
+    }
     // Focus only needs to be set once when the modal is being opened
     if (!this.props.isOpen && newProps.isOpen) {
       this.setFocusAfterRender(true);
+      this.addModalInstance();
```
diasbruno: Perhaps, `this.addModalInstance();` should be called from `this.open();`.

src/components/ModalPortal.js
```diff
@@ -62,16 +69,27 @@ export default class ModalPortal extends Component {
...
       this.open();
     } else if (this.props.isOpen && !newProps.isOpen) {
+      this.removeModalInstance();
```
diasbruno: Perhaps, `this.removeModalInstance();` should be called from `this.close();`.

src/components/ModalPortal.js
```diff
@@ -84,6 +102,7 @@ export default class ModalPortal extends Component {
   }
 
   componentWillUnmount() {
+    this.removeModalInstance();
```
diasbruno: Better let `this.close()` call `this.removeModalInstance();`.

src/components/ModalPortal.js
```diff
@@ -62,16 +69,27 @@ export default class ModalPortal extends Component {
     // Focus needs to be set when mounting and already open
     if (this.props.isOpen) {
       this.setFocusAfterRender(true);
+      this.addModalInstance();
```
diasbruno: Better let `this.open()` call `this.addModalInstance();`.

src/components/Modal.js
```diff
@@ -161,23 +149,9 @@ export default class Modal extends Component {
...
-      elementClass(document.body).add(this.props.bodyOpenClassName);
-    } else {
-      elementClass(document.body).remove(this.props.bodyOpenClassName);
-    }
-
-    if (props.ariaHideApp) {
-      ariaAppHider.toggle(props.isOpen, props.appElement);
-    }
-
     this.portal = renderSubtreeIntoContainer(this, (
```
diasbruno: This is something that bugs me. Not sure what happens with `this.portal` when a modal is open and close too many times. I'll make some tests.
When Modal is unmounted, can this property leaks?

diasbruno:
This is going to bump minor version since we are going to introduce this rule for this class name, so it will be released as `v2.1.x`.

specs/Modal.spec.js
ajfuller: Yeah I'll remove, it was failing during one of my initial attempts, but it's passing now with the latest updates 👍

src/components/ModalPortal.js
ajfuller: I've moved the other ones, but in `componentWillUnmount` we aren't calling the close method at all, so the tests will fail.

We could call `close` method here, but I feel safer calling `removeModalInstance` directly so we don't run into potential issues if a `closeTimeoutMS` prop is set.

src/components/ModalPortal.js
```diff
@@ -99,12 +116,37 @@ export default class ModalPortal extends Component {
     this.content = content;
   }
 
+  addModalInstance() {
```
diasbruno: We can rename `addModalInstance` to `beforeOpen`.

src/components/ModalPortal.js
```diff
@@ -99,12 +116,37 @@ export default class ModalPortal extends Component {
...
+    refCount.add(bodyOpenClassName);
+    // Add body class
+    elementClass(document.body).add(bodyOpenClassName);
+    // Add aria-hidden to appELement
+    if (ariaHideApp) {
+      ariaAppHider.hide(appElement);
+    }
+  }
+
+  removeModalInstance() {
```
diasbruno: `removeModalInstance` can be `beforeClose`.

src/components/ModalPortal.js
diasbruno: We can rename `addModalInstance` to `beforeOpen`.

ajfuller: Changed 👍

diasbruno:
I'm going to make the last release for `v2.0` after merging #441. And later release this patch for version `v2.1`.

diasbruno:
I guess it just needs to be rebase and squash all commits and we can ship it!

src/components/ModalPortal.js
```diff
@@ -99,12 +116,37 @@ export default class ModalPortal extends Component {
     this.content = content;
   }
 
+  beforeOpen() {
+    const { appElement, ariaHideApp, bodyOpenClassName } = this.props;
+    refCount.add(bodyOpenClassName);
+    // Add body class
+    elementClass(document.body).add(bodyOpenClassName);
+    // Add aria-hidden to appELement
```
AndersDJohnson: @ajfuller `appELement` with a big L!

ajfuller:
Screwed up my rebase and accidentally closed, hopefully should be good now

diasbruno:
Just add `[change]` on the message and we can merge.

ajfuller:
@diasbruno I've updated both PR and commit name to include [change]

diasbruno:
Thank you, @ajfuller.

diasbruno:
Great job. Released `v2.1.0`.

PR #447, opened 2017-06-29
[fixed] Modal.removePortal not called when using closetimeoutMS

seanich:
Acceptance Checklist:

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -140,7 +140,7 @@ export default class Modal extends Component {
         this.portal.closeWithTimeout();
       }
 
-      setTimeout(() => this.removePortal, closesAt - now);
+      setTimeout(this.removePortal, closesAt - now);
     } else {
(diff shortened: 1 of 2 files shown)


src/components/Modal.js
```diff
@@ -140,7 +140,7 @@ export default class Modal extends Component {
         this.portal.closeWithTimeout();
       }
 
-      setTimeout(() => this.removePortal, closesAt - now);
+      setTimeout(() => this.removePortal(), closesAt - now);
```
diasbruno: That is correct. You can remove the arrow function and it should work.

`setTimeout(this.removePortal, closesAt - now);`

seanich: Good point, I've updated my commit to remove the arrow function.

diasbruno:
Thank you, @seanich!

diasbruno:
Release `v2.2.1`.

Issue #369, opened 2017-04-11
background element under overlay is draggable when modal is opened in ios safari


chuanxie:
### Summary:
in ios safari, you can drag the background elements under overlay after open modal.
also, if you have a background image on body, even used `position: fixed` or `overflow: hidden`, you still can drag the image

### Steps to reproduce:

 1.  open modal in ios safari
 2. drag around the overlay
 3. background divs is moving and scrolling

### Expected behavior:
should work as desktop and stop scrolling

Aaaaaashu:
@diasbruno I also encountered this problem.

My configuration:
```
<ReactModal
  isOpen={isOpen}
  contentLabel="Modal"
  style={{
    overlay: {
      zIndex: 10,
      backgroundColor: 'rgba(0, 0, 0, 0.5)',
    },
    content: {
      padding: '0',
      borderRadius: '8px',
      border: 'none',
      top: '20px',
      left: '20px',
      right: '20px',
      bottom: 'initial',
      overflow: 'auto',
      WebkitOverflowScrolling: 'touch',
    },
  }}
>
```

musicalshore:
I, too, am encountering this problem.

diasbruno:
@Aaaaaashu @musicalshore Did you find a css trick to prevent this from happening?

musicalshore:
Yes, add `position: fixed` to the body element when the modal is open

indiesquidge:
I'd like to point out that this is a [known issue](https://bugs.webkit.org/show_bug.cgi?id=153852) with Webkit itself, and not something this library has a lot of control over. This also means it will affect any browser using the Webkit engine, including Chrome, Safari, and [others](http://trac.webkit.org/wiki/Applications%20using%20WebKit).

Strangely (at least for me), the issue is only apparent on native iOS devices—there is no issue in the dev tools simulator (tested on Chrome).

The `position: fixed` trick resolves the issue of having the body content scrollable when the modal is open for mobile iOS, but it may introduce new bugs in the process. One that affected me was not having the body element's width set to 100%, so my content was "squished" left whenever a modal was opened, which looked pretty strange.

I am unsure if there are other bugs introduced with this workaround, and perhaps that was just shame-on-me for not having my body width defined, but I thought I'd throw it out there in case other people share the same issue.

(Is it just me who feels like we're [repeating history](https://github.com/twbs/bootstrap/issues/9855) with some of this? Haha.)

wgottschalk:
[...]

PR #452, opened 2017-07-04
[added] Support using multiple document.body classes

### LINKED ISSUE #451: Ever-growing classList when `bodyOpenClassName` prop contains multiple classes
### Summary:

Adding multiple classes to `bodyOpenClassName` will create an ever-growing list of the classes specified.

### Steps to reproduce:

 1. set `bodyOpenClassName` to a string of multiple, space-delineated classes
 2. open and close the modal multiple times
 3. log `document.body.classList` in the console and see that it has continually added the classes but never removed them

### Expected behavior:

All classes added in `bodyOpenClassName` should be removed when the component closes/unmounts.

### Link to example of issue:

https://codepen.io/indiesquidge/pen/VWxzGN

indiesquidge:
Fixes #451.

### Changes proposed:
- update `bodyOpenClassName` prop to handle adding and removing multiple class names
- update String#includes [polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) to work properly
- remove unmaintained and obsolete `element-class` library
- add new helper for adding/removing classes from the body element
- update refCount helper to have a getter and chainable methods

### Use Case

I am using the [Tachyons](http://tachyons.io/) library for CSS styling in my app. If you have not worked with Tachyons before (it's wonderful), it is essentially a collection of atomic class names—i.e., class names that do only one thing (e.g. `.fixed { position: fixed; }`). Because of this, it is in the nature of Tachyons that multiple classes will be tacked on to a single element to achieve a desired style.

Recently, I ran into the issue outlined in #369. To solve the issue, I need to apply two CSS properties to the `document.body` element when the modal opens: `position: fixed` and `overflow: hidden`.

To do this with Tachyons, I need to have `bodyOpenClassName='fixed overflow-hidden'`.

However, this causes a bug in the current version of the `react-modal` library in which the classes are appended ad infinitum ([see here](https://codepen.io/indiesquidge/pen/VWxzGN)).

### Acceptance Checklist:

### DIFF
--- a/docs/styles/classes.md
+++ b/docs/styles/classes.md
@@ -1,10 +1,22 @@
 ### CSS Classes
 
-Sometimes it may be preferable to use CSS classes rather than inline styles.  You can use the `className` and `overlayClassName` props to specify a given CSS class for each of those.
-You can override the default class that is added to `document.body` when the modal is open by defining a property `bodyOpenClassName`.
+Sometimes it may be preferable to use CSS classes rather than inline styles.
 
-It's required that `bodyOpenClassName` must be `constant string`, otherwise we would end up with a complex system to manage which class name
-should appear or be removed from `document.body` from which modal (if using multiple modals simultaneously).
+You can use the `className` and `overlayClassName` props to specify a given CSS
+class for each of those.
 
-Note: If you provide those props all default styles will not be applied, leaving all styles under control of the CSS class.
+You can override the default class that is added to `document.body` when the
+modal is open by defining a property `bodyOpenClassName`.
+
+It's required that `bodyOpenClassName` must be `constant string`, otherwise we
+would end up with a complex system to manage which class name should appear or
+be removed from `document.body` from which modal (if using multiple modals
+simultaneously).
+
+`bodyOpenClassName` can support adding multiple classes to `document.body` when
+the modal is open. Add as many class names as you desire, delineated by spaces.
+
+Note: If you provide those props all default styles will not be applied, leaving
+all styles under control of the CSS class.
+
 The `portalClassName` can also be used however there are no styles by default applied
--- a/specs/Modal.spec.js
+++ b/specs/Modal.spec.js
@@ -271,6 +271,34 @@ describe('State', () => {
     expect(!isBodyWithReactModalOpenClass()).toBeTruthy();
   });
 
+  it('supports adding/removing multiple document.body classes', () => {
+    renderModal({
+      isOpen: true,
+      bodyOpenClassName: 'A B C'
+    });
+    expect(document.body.classList.contains('A', 'B', 'C')).toBeTruthy();
+    unmountModal();
+    expect(!document.body.classList.contains('A', 'B', 'C')).toBeTruthy();
+  });
+
+  it('does not remove shared classes if more than one modal is open', () => {
+    renderModal({
+      isOpen: true,
+      bodyOpenClassName: 'A'
+    });
+    renderModal({
+      isOpen: true,
+      bodyOpenClassName: 'A B'
+    });
+
+    expect(isBodyWithReactModalOpenClass('A B')).toBeTruthy();
+    unmountModal();
+    expect(!isBodyWithReactModalOpenClass('A B')).toBeTruthy();
+    expect(isBodyWithReactModalOpenClass('A')).toBeTruthy();
+    unmountModal();
+    expect(!isBodyWithReactModalOpenClass('A')).toBeTruthy();
+  });
+
   it('should not add classes to document.body for unopened modals', () => {
     renderModal({ isOpen: true });
     expect(isBodyWithReactModalOpenClass()).toBeTruthy();
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -1,10 +1,10 @@
 import React, { Component } from 'react';
 import { PropTypes } from 'prop-types';
-import elementClass from 'element-class';
 import * as focusManager from '../helpers/focusManager';
 import scopeTab from '../helpers/scopeTab';
 import * as ariaAppHider from '../helpers/ariaAppHider';
 import * as refCount from '../helpers/refCount';
+import * as bodyClassList from '../helpers/bodyClassList';
 import SafeHTMLElement from '../helpers/safeHTMLElement';
 
 // so that our CSS is statically analyzable
@@ -119,9 +119,8 @@ export default class ModalPortal extends Component {
 
   beforeOpen() {
     const { appElement, ariaHideApp, bodyOpenClassName } = this.props;
-    refCount.add(bodyOpenClassName);
     // Add body class
-    elementClass(document.body).add(bodyOpenClassName);
+    bodyClassList.add(bodyOpenClassName);
     // Add aria-hidden to appElement
     if (ariaHideApp) {
       ariaAppHider.hide(appElement);
@@ -130,11 +129,8 @@ export default class ModalPortal extends Component {
 
   beforeClose() {
     const { appElement, ariaHideApp, bodyOpenClassName } = this.props;
-    refCount.remove(bodyOpenClassName);
     // Remove class if no more modals are open
-    if (refCount.count(bodyOpenClassName) === 0) {
-      elementClass(document.body).remove(bodyOpenClassName);
-    }
+    bodyClassList.remove(bodyOpenClassName);
     // Reset aria-hidden attribute if all modals have been removed
     if (ariaHideApp && refCount.totalCount() < 1) {
       ariaAppHider.show(appElement);
--- a/package.json
+++ b/package.json
@@ -61,7 +61,6 @@
     "webpack-dev-server": "1.11.0"
   },
   "dependencies": {
-    "element-class": "^0.2.0",
     "exenv": "1.2.0",
     "prop-types": "^15.5.10",
     "react-dom-factories": "^1.0.0"
--- a/specs/helper.js
+++ b/specs/helper.js
@@ -8,9 +8,17 @@ const divStack = [];
 /**
  * Polyfill for String.includes on some node versions.
  */
-if (!(String.prototype.hasOwnProperty('includes'))) {
-  String.prototype.includes = function(item) {
-    return this.length > 0 && this.split(" ").indexOf(item) !== -1;
+if (!String.prototype.includes) {
+  String.prototype.includes = function(search, start) {
+    if (typeof start !== 'number') {
+      start = 0;
+    }
+
+    if (start + search.length > this.length) {
+      return false;
+    }
+
+    return this.indexOf(search, start) !== -1;
   };
 }
 
--- a/src/helpers/bodyClassList.js
+++ b/src/helpers/bodyClassList.js
@@ -1,0 +1,20 @@
+import * as refCount from './refCount';
+
+export function add (bodyClass) {
+  // Increment class(es) on refCount tracker and add class(es) to body
+  bodyClass
+    .split(' ')
+    .map(refCount.add)
+    .forEach(className => document.body.classList.add(className));
+}
+
+export function remove (bodyClass) {
+  const classListMap = refCount.get();
+  // Decrement class(es) from the refCount tracker
+  // and remove unused class(es) from body
+  bodyClass
+    .split(' ')
+    .map(refCount.remove)
+    .filter(className => classListMap[className] === 0)
+    .forEach(className => document.body.classList.remove(className));
+}
--- a/src/helpers/refCount.js
+++ b/src/helpers/refCount.js
@@ -1,23 +1,26 @@
-const modals = {};
+const classListMap = {};
+
+export function get() {
+  return classListMap;
+}
 
 export function add(bodyClass) {
   // Set variable and default if none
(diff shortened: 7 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -30,8 +30,6 @@ always be in sync.
 
 - `npm start` runs the dev server to run/develop examples
 - `npm test` will run the tests.
-- `scripts/test` same as `npm test` but keeps karma running and watches
```
indiesquidge: I removed this line because it seems like it's obsolete. When I run `npm test`, I karma keeps running and watches my changes.

docs/styles/classes.md
```diff
@@ -1,10 +1,21 @@
...
+You can override the default class that is added to `document.body` when the
+modal is open by defining a property `bodyOpenClassName`.
+
+It's required that `bodyOpenClassName` must be `constant string`, otherwise we
+would end up with a complex system to manage which class name should appear or
+be removed from `document.body` from which modal (if using multiple modals
+simultaneously).
+
+`bodyOpenClassName` can support adding multiple classes to `document.body` when
+the modal is open. Add as many class names as you desire, delineated by spaces.
```
indiesquidge: This section

> `bodyOpenClassName` can support adding multiple classes to `document.body` when
> the modal is open. Add as many class names as you desire, delineated by spaces.

is the only newly added documentation. The rest of the changes in this file were simply refactors of the existing documentation into a more easily editable form (e.g. line breaks at ~80 characters, newlines between separate paragraphs, [removal of duplicate spacing](https://github.com/reactjs/react-modal/pull/452/files#diff-80a7f6b1e8f2f0a7f4b86e5ebca280a4L3), etc.)

If you'd prefer I undo these changes and merely add my documentation, just let me know.

diasbruno:
Hi @indiesquidge, the reason for `bodyOpenClassName` is to customize the default `ReactModal__Body--open` and to provide an identifier or group identifier to know when to add or remove the class from `document.body` (in case of multiple modals opened - see `refCount.js`). 

Initially, we made it a single constant class name, so we don't end up with a complex system to manage these classes on the `refCount`.

To accept this PR, this is the property it must hold:

`Modal A`: { open, bodyOpenClassName: `A B` }
`Modal B`: { open, bodyOpenClassName: `A` }

- if `Modal A` is closed, `A` must be in `classList`.
- if `Modal B` is closed, `A B` must be in `classList`.

If it fails on this, one shared class might disappear and it would be hard to find why.

The same result can be achieved using `onAfterOpen` and `onRequestClose` to inject and remove additional class names to the `document.body`.

CONTRIBUTING.md
diasbruno: This can be moved to another PR.

src/components/ModalPortal.js
```diff
@@ -131,9 +132,11 @@ export default class ModalPortal extends Component {
   beforeClose() {
     const { appElement, ariaHideApp, bodyOpenClassName } = this.props;
     refCount.remove(bodyOpenClassName);
-    // Remove class if no more modals are open
+    // Remove class(es) if no more modals are open
     if (refCount.count(bodyOpenClassName) === 0) {
-      elementClass(document.body).remove(bodyOpenClassName);
+      bodyOpenClassName.split(' ').forEach(className => {
+        document.body.classList.remove(className);
+      });
```
diasbruno: This might remove a shared class if more than one modal is opened.

indiesquidge:
@diasbruno, thanks for the tips! I have updated my code to handle the test case you laid out about modals using shared classes.

> the reason for bodyOpenClassName is to customize the default ReactModal__Body--open and to provide an identifier or group identifier to know when to add or remove the class from document.body

This makes sense for sure. I'm just hoping to be able to add multiple classes to `document.body` via the `bodyOpenClassName` prop, similar to how I can have a string of classnames for `className` and `overlayClassName`.

> Initially, we made it a single constant class name, so we don't end up with a complex system to manage these classes on the refCount

I definitely agree that a complex system to manage the classes would be bad. I see that `refCount` is really just managing a mapping of classnames (keys) to a number of how many modals "added" each class (values). I took advantage of mapping in my update, with a few slight tweaks. Hopefully what I have implemented is not too complex.

> The same result can be achieved using onAfterOpen and onRequestClose to inject and remove additional class names to the document.body.

Using these hooks is what I initially looked at doing, but I'm already doing a number of other things in each of the methods and figured having the ability to add multiple classes via the `bodyOpenClassName` prop would be worthwhile feature to have.

diasbruno:
You can remove the artifacts (`dist/*`). It's generated when publishing the new version.
Later I'll finish to review this PR.

indiesquidge:
Tests seem to be breaking on Node 6 & 7, but it works fine on my local machine (Node 8). Not sure what the issue is; I'll look into it later.

indiesquidge:
Strange, even when I switch my node version to 6.11.0, all tests still pass on my machine...

diasbruno:
Yeah, it seems to be working (I've download your branch and run the tests on `node v8.1.2`).
Note that `travis-ci` runs on `Firefox 38.0.0`.

specs/Modal.spec.js
```diff
@@ -271,6 +271,34 @@ describe('State', () => {
...
+      isOpen: true,
+      bodyOpenClassName: 'A B'
+    });
+
+    expect(isBodyWithReactModalOpenClass('A B')).toBe(true);
+    unmountModal();
+    expect(isBodyWithReactModalOpenClass('A B')).toBe(false);
+    expect(isBodyWithReactModalOpenClass('A')).toBe(true);
+    unmountModal();
+    expect(isBodyWithReactModalOpenClass('A')).toBe(false);
```
diasbruno: Expectations should be always true.

```js
expect(!isBodyWithReactModalOpenClass('A B')).toBeTruth();
expect(isBodyWithReactModalOpenClass('A')).toBeTruthy();
```

src/components/ModalPortal.js
```diff
@@ -119,9 +118,11 @@ export default class ModalPortal extends Component {
...
   beforeOpen() {
     const { appElement, ariaHideApp, bodyOpenClassName } = this.props;
-    refCount.add(bodyOpenClassName);
-    // Add body class
-    elementClass(document.body).add(bodyOpenClassName);
+    // Increment class(es) on refCount tracker and add class(es) to body
+    bodyOpenClassName
+      .split(' ')
+      .map(refCount.add)
+      .forEach(bodyClass => document.body.classList.add(bodyClass));
```
diasbruno: You can make this pipeline a function.

src/components/ModalPortal.js
```diff
@@ -130,11 +131,14 @@ export default class ModalPortal extends Component {
...
-      elementClass(document.body).remove(bodyOpenClassName);
-    }
+    const classListMap = refCount.get();
+    // Decrement class(es) from the refCount tracker
+    // and remove unused class(es) from body
+    bodyOpenClassName
+      .split(' ')
+      .map(refCount.remove)
+      .filter(bodyClass => classListMap[bodyClass] === 0)
+      .forEach(bodyClass => document.body.classList.remove(bodyClass));
```
diasbruno: You can make this pipeline a function.

specs/Modal.spec.js
indiesquidge: Sure, I can certainly change the the tests to use the `toBeTruthy()` method, but may I ask why it is preferred?

While working with these tests, the terseness of using `toBeTruthy()` on bang methods methods—instead of something like `toBe(false)`—made it more difficult to see what I was asserting.

src/components/ModalPortal.js
indiesquidge: Something like this okay?

```javascript
  addClassToBody = (bodyClass) => {
    // Increment class(es) on refCount tracker and add class(es) to body
    bodyClass
      .split(' ')
      .map(refCount.add)
      .forEach(className => document.body.classList.add(className));
  }
```

src/components/ModalPortal.js
indiesquidge: Something like this okay?

```javascript
  removeClassFromBody = (bodyClass) => {
    const classListMap = refCount.get();
    // Decrement class(es) from the refCount tracker
    // and remove unused class(es) from body
    bodyClass
      .split(' ')
      .map(refCount.remove)
      .filter(className => classListMap[className] === 0)
      .forEach(className => document.body.classList.remove(className));
  }
```

specs/Modal.spec.js
diasbruno: Just to follow the current code style.

My opinion: I'd like to see something like `expect(!something_that_is_false);`, because every expectation ends up been a true statement.

`F = F => T`

src/components/ModalPortal.js
diasbruno: I think you can create a new file on helpers folder.

indiesquidge: Sweet! Any preference on the name? I was thinking something like `bodyManager` or `bodyElementManager`, but am definitely open to other suggestions.

src/components/ModalPortal.js
```diff
@@ -119,9 +119,8 @@ export default class ModalPortal extends Component {
 
   beforeOpen() {
     const { appElement, ariaHideApp, bodyOpenClassName } = this.props;
-    refCount.add(bodyOpenClassName);
     // Add body class
-    elementClass(document.body).add(bodyOpenClassName);
+    bodyClassList.add(bodyOpenClassName);
```
indiesquidge: As this is iterated upon, it seems like we are moving in a direction very similar to the `elementClass` library that was here before. (The major under-the-hood difference being that `bodyOpenClassName` can be a string of multiple class names, and using the other helper `refCount` to keep track of opened modals relying on the class internally.)

If you would prefer, I could make `bodyClassList` look and feel similar to `elementClass`, i.e. have `document.body` be passed in as an argument like it is here. This would remove the line change here, thus keeping the "public API" of the helper class the same as how `elementClass` looked.

indiesquidge: I've created a [commit](https://github.com/indiesquidge/react-modal/commit/6b83e2609335720548085f5d7fc20fcc973b4d34#diff-eafaf565940b2fb71c5ed190e4f65abf) on another branch to demonstrate what this might look like.

indiesquidge:
> Note that travis-ci uses Firefox 38.0.0 to run the tests.

@diasbruno, is there any way to simulate this test run locally? Setting my node version to the same still passes for me locally; running the exact command Travis does (i.e. `node_modules/.bin/cross-env NODE_ENV=test node_modules/.bin/karma start "--single-run"`) still passes for me locally.

Basically anything I can think of to try and "break" the test script locally by mimicking the Travis build is not working—the tests always seem to behave as I would expect when I run them locally.

diasbruno:
The implementation for this is ok. You can try installing Firefox 38 and run `CONTINUOUS_INTEGRATION=true make tests-single-run`.

diasbruno:
ping @ajfuller. Any thoughts?

src/components/ModalPortal.js
indiesquidge: @diasbruno, which implementation would you prefer? Does the thumbs up mean you think I should implement the code outlined in the commit I linked to? I think the code in the external commit is a more robust solution than what currently exists in this PR, but it's up to you :)

diasbruno: ow sorry. the version that uses `document.body` directly. :)

ajfuller:
Code looks nice to me 👍

I don't see anything that stands out as to why it's failing on Firefox 38 either. I'll try to pull down and test later, but here is the link to download Firefox 38:
https://ftp.mozilla.org/pub/firefox/releases/38.8.0esr/mac/en-US/

diasbruno:
@indiesquidge I think you can push a commit with `console.log` on each step of the test to see what's going on.

indiesquidge:
@diasbruno, alright I finally got the test failing locally (thanks, @ajfuller for the link to Firefox 38).

I narrowed the issue down to String#includes. Firefox 38 does not have this function natively, so it relies on the polyfill provided in `specs/helper.js`. However, this polyfill doesn't appear to work correctly when given a string like `'A B'` because it will split that string on the space and then check it against the unsplit string, which of course fails.

The polyfill breaks on other tests that should return true

```javascript
String.prototype.includes = function(item) {
	return this.length > 0 && this.split(" ").indexOf(item) !== -1;
}

const str = 'To be, or not to be, that is the question.'
console.log(str.includes('To be')) // returns false; should be true
```

I have updated the polyfill to use the [MDN polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes), which seems to work properly and indeed the test passes after this change.

diasbruno:
@indiesquidge This is great! Sorry for the long code review and thanks for the effort on this feature.

diasbruno:
@indiesquidge I'm merging this PR. Maybe later, we can think on how to remove the polyfill. Maybe we can run those tests with `babel-polyfill` (?) or something.

indiesquidge:
Thanks @diasbruno for speedy replies and for helping me along with this PR. Glad I could help :)

diasbruno:
Released `v2.2.2`.

PR #480, opened 2017-08-16
[fixed] Remove required flag from contentLabel propType in Modal

### LINKED ISSUE #479: Make contentLabel optional
According to the ARIA spec, "If the label text is visible on screen, authors SHOULD use aria-labelledby and SHOULD NOT use aria-label." With that said, since the contentLabel uses aria-label, it shouldn't be required since ideally, if the label is visible on the page, then an id to that element should be provided to aria-labelledby.

https://www.w3.org/TR/wai-aria/states_and_properties#aria-label

bluSCALE4:
Fixes #479.

Changes proposed:
ARIA spec, "If the label text is visible on screen, authors SHOULD use aria-labelledby and SHOULD NOT use aria-label." 
Changing contentLabel propType to not be required; reference id for aria-labelledby will now be supported.

Won't affect existing code.

Acceptance Checklist:
- [X] All commits have been squashed to one.
- [X] If the commit message has [changed] or [removed], there is an upgrade path above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,10 +27,9 @@ To install the stable version you can use [npm](https://npmjs.org/) or [yarn](ht
 
 ## Usage
 
-The Modal object has two required props:
+The Modal object has one required prop:
 
 - `isOpen` to render its children.
-- `contentLabel` to improve accessibility, since `v1.6.0`.
 
 Example:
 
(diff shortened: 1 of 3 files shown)


diasbruno:
Hi @bluSCALE4, can you check if it is needed to update the docs?

README.md
```diff
@@ -1,4 +1,4 @@
-# react-modal
+k react-modal
```
diasbruno: Use "#".

bluSCALE4:
@diasbruno sorry about that.

diasbruno:
Oh, my bad. Totally forgot this PR. Sorry, @bluSCALE4. I'll get this merged soon.

PR #490, opened 2017-09-01
[added] Don't focus after render if we don't want to

### LINKED ISSUE #306: Ability to disable/delay automatic focus
### Summary:
It would be beneficial to have another prop on <Modal> to disable or delay the automatic focus.

Specifically on iOS Safari, the page is automatically scrolled to the focused object. There does not appear to be any way to suppress this functionality, and workarounds are quirky. If movement transitions are applied to the modal, the screen scrolls to undesired locations. Being able to disable or set a specific delay on the focus would provide an easy workaround.

imadha:
Fixes #306 .

Changes proposed:
- Adds a `shouldFocusAfterRender` prop, defaults to `true`

 - No upgrade necessary

Acceptance Checklist:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -38,7 +38,7 @@ import ReactModal from 'react-modal';
   */
   closeTimeoutMS={0}
   /*
-    Object indicating styles to be used for the modal.  
+    Object indicating styles to be used for the modal.
     It has two keys, `overlay` and `content`.  See the `Styles` section for more details.
   */
   style={{ overlay: {}, content: {} }}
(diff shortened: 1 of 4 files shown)


diasbruno:
Hi @imadha, can you update the documentation for this feature?

diasbruno:
Forget the previous comment. Everything is ok.

diasbruno:
Released `v2.3.1`. Thank you @imadha. 

I think we need to update the documentation telling that it's not recommended to disable the focus due to some problems with accessibility and, perhaps, should be only used in case of using `react-modal` as a base component for other components. Any thoughts?

Issue #484, opened 2017-08-23
React Fiber support


maschwenk:
It appears that this library is not compatible with React Fiber. It is using some deprecated/removed React API's. Are there any plans in place to get start this work? Curious how much work is involved. 

https://github.com/facebook/react/issues/10294#issuecomment-318210497

diasbruno:
@maschwenk Yeah, that's for sure. Thanks for the link and PRs are always welcome.

maschwenk:
Initial investigation shows that we need to replace `unstable_renderSubtreeIntoContainer` with `unstable_createPortal`. React maintainer details it here: https://github.com/facebook/react/issues/9808#issuecomment-308085984 and provided this fiddle https://jsfiddle.net/r442t0mk/

sandys:
we are also getting impacted by this issue.

diasbruno:
@maschwenk @sandys I'll merge this issue with @amannn's issue #498 to tackle this. Perhaps, we can open a project and make a roadmap for the next version.

diasbruno:
`react-modal@3.0.0-alpha` is available. 

Version `2.3.3` is the stable version and can be installed with `react-modal@latest`.

To use the compatible version with React 16, you can install by running `npm install react-modal@next`.

merlinpatt:
Why use `unstable_createPortal`? Why not use `ReactDOM.createPortal`?

https://egghead.io/lessons/react-render-elements-outside-the-current-react-tree-using-portals-in-react-16

diasbruno:
@merlinpatt I think this was before `createPortal` got in the API.

diasbruno:
Released official v3.0.0. 

You can follow #498 for more information.

Thank you all.

PR #512, opened 2017-09-28
WIP: [feature] update react-modal to work with React 16.

### LINKED ISSUE #498: React 16 compatibility
### Summary:

I looked around in the issue tracker but couldn't find an existing issue. As React 16 is right around the corner, maybe it's time to prepare a PR that adds support for it?

It looks like at least the [call to `unstable_renderSubtreeIntoContainer`](https://github.com/reactjs/react-modal/blob/f007aeb7336bc8a47bb6b8c7919ed3a077fbd87d/src/components/Modal.js#L11) should be updated as there's now first class support for portals.

### Steps to reproduce:

 1. Render a modal.
 2. Unmount it.

### Expected behavior:

Shouldn't throw any errors.

### Link to example of issue:

https://codesandbox.io/s/xpl8r853o4

diasbruno:
This update is based on the implementation on [CogniStreamer/react-modal@4434c4a](https://github.com/CogniStreamer/react-modal/commit/4434c4aac57d91a24d8f1b89e85a41fe2bd2023d).

It's still WIP, need to update documentation.

Fixes #498.
Related to #484.

Changes proposed:
- Use the correct way to work with portal in react 16.

- Update your project to use react 16.

Acceptance Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,8 +46,8 @@
     "karma-webpack": "^2.0.4",
     "mocha": "3.5.3",
     "npm-run-all": "^4.1.1",
-    "react": "^15.6.1",
-    "react-dom": "^15.6.1",
+    "react": "^16",
+    "react-dom": "^16",
     "react-router": "^4.2.0",
     "react-router-dom": "^4.2.2",
     "should": "^13.1.0",
@@ -62,8 +62,8 @@
     "react-dom-factories": "^1.0.0"
   },
   "peerDependencies": {
-    "react": "^0.14.0 || ^15.0.0",
-    "react-dom": "^0.14.0 || ^15.0.0"
+    "react": "^16.0.0",
+    "react-dom": "^16.0.0"
   },
   "tags": [
     "react",
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -8,13 +8,16 @@ import SafeHTMLElement from '../helpers/safeHTMLElement';
 export const portalClassName = 'ReactModalPortal';
 export const bodyOpenClassName = 'ReactModal__Body--open';
 
-const renderSubtreeIntoContainer = ReactDOM.unstable_renderSubtreeIntoContainer;
-
 function getParentElement(parentSelector) {
   return parentSelector();
 }
 
 export default class Modal extends Component {
+  constructor(props) {
+    super(props);
+    this.node = document.createElement('div');
+  }
+
   static setAppElement(element) {
     ariaAppHider.setElement(element);
   }
@@ -97,13 +100,10 @@ export default class Modal extends Component {
   };
 
   componentDidMount() {
-    this.node = document.createElement('div');
     this.node.className = this.props.portalClassName;
 
     const parent = getParentElement(this.props.parentSelector);
     parent.appendChild(this.node);
-
-    this.renderPortal(this.props);
   }
 
   componentWillReceiveProps(newProps) {
@@ -118,8 +118,6 @@ export default class Modal extends Component {
       currentParent.removeChild(this.node);
       newParent.appendChild(this.node);
     }
-
-    this.renderPortal(newProps);
   }
 
   componentWillUpdate(newProps) {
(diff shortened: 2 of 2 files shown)


diasbruno:
I didn't remove `unmountComponentAtNode(node)` yet, because the render method of `Modal` will have to change to return `null`.

bjrmatos:
is still rendering normally when doing server side rendering? i think that returning `null` was the thing that make possible to not break the component when doing SSR.

bjrmatos:
also doing `this.node = document.createElement('div');` in `componentDidMount` was other of the reasons, moving to the constructor breaks SSR.

diasbruno:
@bjrmatos Thanks. Can you PR these changes to `diasbruno/react-moda:feature/react-16`?

catamphetamine:
Current changes do break server side rendering as @bjrmatos says.
I'd leave `this.node = ...` code where it was, in `componentDidMount` because it's the only place where `document` exists.
And then in `componentDidMount` it would just `this.setState({ node: this.node })` to rerender (or just `this.forceUpdate()`).

package.json
```diff
@@ -62,8 +62,8 @@
     "react-dom-factories": "^1.0.0"
   },
   "peerDependencies": {
-    "react": "^0.14.0 || ^15.0.0",
-    "react-dom": "^0.14.0 || ^15.0.0"
+    "react": "^16.0.0",
+    "react-dom": "^16.0.0"
```
emilyrohrbough: Are you removing compatibility for `^0.14.0 || ^15.0.0`?

catamphetamine:
> @emilyrohrbough Are you removing compatibility for `^0.14.0 || ^15.0.0`?

Yeah, that's also a controversial decision made in the changes.
It should support all Reacts instead, feature-detecting the `ReactDOM.createPortal()` method existence and then behaving the old way in case of older React and behaving the new way in case of a newer React.

package.json
diasbruno: Not really, it's just a initial PR :)

diasbruno:
@catamphetamine @emilyrohrbough This PR is available for everyone to participate, give some ideas, etc...It still WIP, as you can see in the title.

catamphetamine:
@diasbruno I took your ideas further and introduced a feature flag so that it's compatible with both old React and new React.
I didn't test the code though.

package.json
diasbruno: I'm forcing this temporary just to test what would need to change in regarding to support react 16. If possible, I'll guarantee backwards compatibility with previous version of react.

diasbruno:
@catamphetamine Thanks. You can PR you changes to my branch. So we can keep all in one place.

catamphetamine:
@diasbruno "You may not have permission to access react-modal. Check Preferences to make sure you’re still logged in."

If you want my changes in this branch then you should add me as a collaborator in the settings

diasbruno:
@catamphetamine I'll close this issue in favor of #513.

andrepcg:
Using react-modal@next with Next.js to server side render a react app I'm getting

image

`document is not defined
ReferenceError: document is not defined
    at Modal.render (/Users/aper/DEV/backen/node_modules/react-modal/lib/components/Modal.js:150:21)`

diasbruno:
@andrepcg Please, you can open a issue. 

Probably checking for `typeof window !== undefined` is not enough.

catamphetamine:
@diasbruno Perhaps you should check `document` instead of `window` because it makes more sense since you're not using `window`

catamphetamine:
@diasbruno MAAAAN
WTF
```js
typeof window !== undefined
```
No no

```js
typeof window !== 'undefined'
```

catamphetamine:
@andrepcg Don't open an issue, the bug is found

diasbruno:
@catamphetamine Oh my. This kind of thing happen at 3am.

diasbruno:
I'll fix it and release a new version. Thanks, @catamphetamine.

diasbruno:
@catamphetamine @andrepcg #516 should be a safer check for `canUseDOM`. Can you review it, please?

PR #513, opened 2017-09-28
React16

### LINKED ISSUE #498: React 16 compatibility

catamphetamine:
Fixes #498

Based on the implementation on [CogniStreamer/react-modal@4434c4a](https://github.com/CogniStreamer/react-modal/commit/4434c4aac57d91a24d8f1b89e85a41fe2bd2023d) 
See all details there.

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -159,7 +159,9 @@ export default class Modal extends Component {
   portalRef = (portal) => this.portal = portal;
 
   removePortal = () => {
-    ReactDOM.unmountComponentAtNode(this.node);
+    if (!useReactPortal) {
+      ReactDOM.unmountComponentAtNode(this.node);
+    }
     const parent = getParentElement(this.props.parentSelector);
     parent.removeChild(this.node);
   }
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -8,6 +8,8 @@ import SafeHTMLElement from '../helpers/safeHTMLElement';
 export const portalClassName = 'ReactModalPortal';
 export const bodyOpenClassName = 'ReactModal__Body--open';
 
+const useReactPortal = ReactDOM.createPortal !== undefined;
+
 const renderSubtreeIntoContainer = ReactDOM.unstable_renderSubtreeIntoContainer;
 
 function getParentElement(parentSelector) {
@@ -103,7 +105,11 @@ export default class Modal extends Component {
     const parent = getParentElement(this.props.parentSelector);
     parent.appendChild(this.node);
 
-    this.renderPortal(this.props);
+    if (useReactPortal) {
+      this.forceUpdate();
+    } else {
+      this.renderPortal(this.props);
+    }
   }
 
   componentWillReceiveProps(newProps) {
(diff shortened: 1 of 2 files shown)


src/components/Modal.js
```diff
@@ -8,6 +8,8 @@ import SafeHTMLElement from '../helpers/safeHTMLElement';
 export const portalClassName = 'ReactModalPortal';
 export const bodyOpenClassName = 'ReactModal__Body--open';
 
+const useReactPortal = ReactDOM.createPortal !== undefined;
```
nickhudkins: Should we introduce a warning here indicating that usage without ReactDOM.createPortal is now deprecated and will be removed in the next major release?

catamphetamine: Then every React 15 user will get those warnings so they'll create issues about disabling those warnings.
Most likely some folks out there are using legacy React 14 even.

diasbruno: @catamphetamine @nickhudkins Remember that all the changes to support react 16 must target the branch `next`. This way we can safely make any changes.

marlonbernardes: What about upgrading to `3.0.0` and supporting only React 16+?

diasbruno: @marlonbernardes It would be preferable to keep it backwards compatible. Also, to avoid maintaining 2 versions.

nickhudkins:
3.0.0 seems like a good opportunity to do some heavier work (if there is
work to be done) but I think backwards compatible and encouragement for
people to keep moving forward is a good path

src/components/Modal.js
marlonbernardes: It makes perfect sense. Let me know if there's anything I can help you with.

catamphetamine:
I added @diasbruno to collaborators to the repo:
https://github.com/catamphetamine/react-modal
Anyway this PR is gonna be closed I guess because I heard that there will be a special `next` branch for these changes.
If anyone wants to take part committing code then leave a comment here requesting to be added to collaborators.

diasbruno:
@catamphetamine Please, can you change the target branch to be `next`?
 
@nickhudkins Agreed. Once we are done with `next` and we are ready to merge with `master`,  the new release will be `3.0`.

catamphetamine:
@diasbruno Ok, I'm closing this PR and creating a new one then

catamphetamine:
Seems that closing PR is not required to switch branches.
That's good feature.
Switched to `next`.

tomasztomys:
Can you release new version with this PR asap?

catamphetamine:
@tomasztomys This PR was not accepted by the maintainers of the library so I should actually close not to confuse other users.
Consult the developers of this library on the estimated progress on this issue.
I'm no longer involved.

lawlmart:
Sorry if I'm confused, but shouldn't this have been merged into next instead of closed?

catamphetamine:
@lawlmart The developers of this library decided to take some other approach.
https://github.com/catamphetamine/react-modal/pull/1#issuecomment-333341305
Something related to refactoring.

PR #516, opened 2017-10-04
[fixed] typeof returns a string (canUseDOM).

diasbruno:
This happens when testing for `canUseDOM`.

Changes proposed:
-

- None

Acceptance Checklist:

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -8,7 +8,10 @@ import SafeHTMLElement from '../helpers/safeHTMLElement';
 export const portalClassName = 'ReactModalPortal';
 export const bodyOpenClassName = 'ReactModal__Body--open';
 
-const canUseDOM = typeof window !== undefined;
+const canUseDOM = typeof window !== 'undefined' || (
+  typeof document !== 'undefined' && document.createElement
+);
+
 const isReact16 = ReactDOM.createPortal !== undefined;
 const createPortal = isReact16 ?
   ReactDOM.createPortal :


src/components/Modal.js
```diff
@@ -8,7 +8,10 @@ import SafeHTMLElement from '../helpers/safeHTMLElement';
 export const portalClassName = 'ReactModalPortal';
 export const bodyOpenClassName = 'ReactModal__Body--open';
 
-const canUseDOM = typeof window !== undefined;
+const canUseDOM = typeof window !== 'undefined' || (
+  document && document.createElement !== undefined
```
catamphetamine: You have checked for `window` existence but you haven't checked for `document` existence therefore if `window` exists and `document` does not then it will throw.
I'd rewrite that as something like:

```js
const canUseDOM = typeof document !== 'undefined' && typeof document.createElement === 'function'
```

diasbruno:
It should be ok now. Tested on node.

catamphetamine:
@diasbruno 
Still semantially your code is weird

```js
const canUseDOM = typeof window !== 'undefined' || (
  typeof document !== 'undefined' && document.createElement
);
```

So you're demanding either `window` or `document`.
And at the same time you're not using `window` anywhere.
Not to mention that even if `window` exists it doesn't mean `document` exists.

diasbruno:
`window` is just a first bail and if window is present, we go ahead to check for document.

catamphetamine:
> and if window is present, we go ahead to check for document.

No

diasbruno:
Sorry, merge too soon.

diasbruno:
Updated the checker to use your. Thank you.

catamphetamine:
Yeah, looks good now

PR #518, opened 2017-10-05
[fixed] Drag stop (mouseup) on Overlay closes Modal

### LINKED ISSUE #515: Drag stop (mouseup) on Overlay closes Modal
### Summary: 
I'v seen some other issues that was marked as resolved, but I can reproduce this Issue in current version 2.3.3. In case I stop dragging on the overlay, the modal is closing. I have added an example: https://codesandbox.io/s/1r60668vm7

### Steps to reproduce:

 1. Open the modal in the example.
 2. Drag the slider around and move it outside of the modal.
 3. Stop dragging (mouseup) outside of the modal and it will be closed.

### Expected behavior: 
The modal should stay opened.

### Link to example of issue:
https://codesandbox.io/s/1r60668vm7

mrsilver256:
Fixes #515.

Changes proposed:
- Fixed drag stop on overlay closes modal
- Added moveFromContentToOverlay variable to check whether or not users drag from content to overlay

Acceptance Checklist:

### DIFF
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -65,6 +65,7 @@ export default class ModalPortal extends Component {
     };
 
     this.shouldClose = null;
+    this.moveFromContentToOverlay = null;
   }
 
   componentDidMount() {
@@ -216,6 +217,21 @@ export default class ModalPortal extends Component {
       }
     }
     this.shouldClose = null;
+    this.moveFromContentToOverlay = null;
+  }
+
+  handleOverlayOnMouseUp = () => {
+    if (this.moveFromContentToOverlay === null) {
+      this.shouldClose = false;
+    }
+  }
+
+  handleContentOnMouseUp = () => {
+    this.shouldClose = false;
+  }
+
+  handleOverlayOnMouseDown = () => {
+    this.moveFromContentToOverlay = false;
   }
 
   handleContentOnClick = () => {
@@ -224,6 +240,7 @@ export default class ModalPortal extends Component {
 
   handleContentOnMouseDown = () => {
     this.shouldClose = false;
+    this.moveFromContentToOverlay = false;
   }
 
   requestClose = event =>
@@ -271,7 +288,9 @@ export default class ModalPortal extends Component {
         ref={this.setOverlayRef}
         className={this.buildClassName('overlay', overlayClassName)}
         style={{ ...overlayStyles, ...this.props.style.overlay }}
-        onClick={this.handleOverlayOnClick}>
+        onClick={this.handleOverlayOnClick}
+        onMouseDown={this.handleOverlayOnMouseDown}
+        onMouseUp={this.handleOverlayOnMouseUp}>
         <div
           ref={this.setContentRef}
           style={{ ...contentStyles, ...this.props.style.content }}
@@ -279,6 +298,7 @@ export default class ModalPortal extends Component {
           tabIndex="-1"
           onKeyDown={this.handleKeyDown}
           onMouseDown={this.handleContentOnMouseDown}
+          onMouseUp={this.handleContentOnMouseUp}
           onClick={this.handleContentOnClick}
           role={this.props.role}
           aria-label={this.props.contentLabel}


src/components/ModalPortal.js
```diff
@@ -216,6 +217,21 @@ export default class ModalPortal extends Component {
       }
     }
     this.shouldClose = null;
+    this.moveFromContentToOverlay = null;
+  }
+
+  handleOverlayOnMouseUp = () => {
+    if (this.moveFromContentToOverlay === null) {
```
diasbruno: Perhaps we can simplify this by inspecting the `event.target` or `event.currentTarget` is inside the `this.content`.
`this.shouldClose = this.content.contains(event.target)`.

mrsilver256: Nice!
Surely, when the mouse move from content then `mouseUp` on `overlay`, the code `this.content.contains(event.target)` always return `false`. 
Basically, we will need to take more steps to get the false value, however, we can get a better understanding.

diasbruno: Ok. You can use `shouldClose` instead of a new property.

diasbruno: We can make a nice refactoring about all the mouse events handle. Currently, this flag `shouldClose` has 3 states `null | false | true`. We can improve it by moving the logic from `handleOverlayClick` to `handleOverlayMouseUp`. This way only then this event is fired we actually try to close the modal.

`handleOverlayMouseDown` is the tricky part. With this event, we actually set the `shouldClose` to `true` when the `event.target == event.currentTarget`, this way we know that the click was in the `overlay`.

All the other handles must set `shouldClose` to `false`.

diasbruno: Here is the idea.

```js
  handleOverlayOnMouseDown = event => {
    this.shouldClose = event.target == event.currentTarget;
  }

  handleOverlayOnMouseUp = (event) => {
    if (this.shouldClose && this.props.shouldCloseOnOverlayClick) {
      this.shouldClose = false;

      if (this.ownerHandlesClose()) {
        this.requestClose(event);
      } else {
        this.focusContent();
      }
    }
  }
```

mrsilver256: It's very nice to have the tricky way `handleOverlayOnMouseDown` by checking the `event.target` and `event.currentTarget`. Unfortunately, It does not work.
The reason is `handleOverlayOnMouseDown` will not be called when `mouseDown` on the `Slider`, that `Slider` might `stopPropagation` so the `overlay` receives no events.
Here is how I solve the problem:
`handleOverlayOnMouseDown` does not call when `mouseDown` on `Slider`, I init a new prop here which is `moveFromContentToOverlay` with a `null` value, then in `onMouseUp`, if it is still `null` => `this.shouldClose = false`. `null` value here indicates that users `mouseDown` at `slider` then all other cases should set `moveFromContentToOverlay` to `false`. The `moveFromContentToOverlay` is actually another flag, because the `shouldClose` and its 3 flags `(null|false|true)` have been used for your purposes so I need to use another flag. Perhaps there's a better way to solve this problem instead of initialize a new prop, but I find it no way.

diasbruno: @mrsilver256 That's fine. 👍 I'm just trying to avoid handling too much state as possible...but if we can't find a better way, I'll be happy to merge this PR.

Check out this branch, [diasbruno/react-modal:fix/handle-mouse-events](https://github.com/diasbruno/react-modal/tree/fix/handle-mouse-events).

mrsilver256: 👍 Nice!

mrsilver256: @diasbruno Do you know when will this be merged and released?

diasbruno: Before merging this PR, I'm planing to make `v3` official.

mrsilver256: That would take a long time. I'm actually using React Modal inside my project and also get this error, I'd manually handle this, btw, until this get merged.

diasbruno:
I'll merge this since it's a confirmed issue. Later we can refactor and/or improve the implementation.
Thank you @mrsilver256 for the patch and patience.

PR #521, opened 2017-10-06
[fixed] correctly walk when using TAB.

### LINKED ISSUE #511: Tabbing with radio fieldsets
### Steps to reproduce:

 1. Create a modal with a fieldset inside, where the fieldset contains at least two radio inputs and there are no focusable elements after the fieldset in the modal.
 2. Open the modal and attempt using tab to navigate
 3. When focus is within the fieldset, tab will move focus outside the modal. If the modal contained the final focusable element in the DOM, this means the window will blur; if it was not, then focus will move to the next element outside of it.

### Expected behavior:

Focus should stay within the modal.

### Link to example of issue:
Behavior when final focusable element in DOM is in modal: https://codesandbox.io/s/q4lk918jww
Behavior when focusable elements in DOM follows modal: https://codesandbox.io/s/j2jjvljp7w

### Additional notes:

Radio groups have tabbing behavior that is not being accounted for by the tab intercept logic in react-modal. That logic bails and defers to the default behavior except when the active element is the final focusable element in the modal. However, radio inputs within a fieldset have keyboard behavior for accessibility with a bit more nuance. When you tab into a radio group "from above", you land on the first input, and when you tab into a radio group "from below", you land on the last input, yet within the group, you navigate with arrow keys, not tab; instead, tab exits the whole group. So when the first input is focused and you press tab, react-modal sees correctly that it is not the final focusable element, but it assumes this also means that the native tab behavior will be to go to the next input element — which is not true. Natively, the next focusable element in this situation is whatever follows the whole group. So by bailing out at this point, focus is able to escape the modal.

diasbruno:
Fixes #511.

This is still wip, for test only.

Changes proposed:
- fix the logic to walk on "tabbable" elements inside the modal.

- None

Acceptance Checklist:

### DIFF
--- a/examples/basic/app.js
+++ b/examples/basic/app.js
@@ -3,6 +3,7 @@ import ReactDOM from 'react-dom';
 import Modal from 'react-modal';
 import SimpleUsage from './simple_usage';
 import MultipleModals from './multiple_modals';
+import Forms from './forms';
 import ReactRouter from './react-router';
 
 const appElement = document.getElementById('example');
@@ -11,6 +12,7 @@ Modal.setAppElement('#example');
 
 const examples = [
   SimpleUsage,
+  Forms,
   MultipleModals,
   ReactRouter
 ];
--- a/examples/basic/forms/index.js
+++ b/examples/basic/forms/index.js
@@ -1,89 +1,71 @@
 import React, { Component } from 'react';
 import Modal from 'react-modal';
-import MyModal from './modal';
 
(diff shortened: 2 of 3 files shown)


diasbruno:
cc @bathos.

diasbruno:
Today, I'll finish this PR.

diasbruno:
How this stuff got here. Sorry everyone that was notified by this PR.

PR #535, opened 2017-10-16
[fixed] Removes body classNames after the modal is closed.

### LINKED ISSUE #508: Remove body class after modal closes.
### Summary:
I think that the body className removal should be moved to afterClose.  The reasoning being, if you are closing via closeTimeoutMS you probably don't want dom clean up to happen before the modal is actually removed from the dom. For example, i use a 'dialog-open' bodyOpenClassName to remove vertical overflow on the body. When i close the modal it transitions out over 250ms. Since bodyClassList is updated before the modal is actually disposed, it restores scrollbars before the transition happens and makes the entire page look janky until closeTimeoutMS expires and the modal is finally removed.

### Steps to reproduce:
 1. Create a modal with a bodyOpenClassName and closeTimeoutMS
 2. Notice how the bodyOpenClassName is removed before the modal is disposed.

### Expected behavior:
I feel that removing body classNames should happen after the modal has been closed.

### Link to example of issue:
https://codesandbox.io/s/2o22zw3ryy

EduardoNogueira:
Fixes #508.

Changes proposed:
- Remove body classNames only after the modal is closed.

Acceptance Checklist:
- [ ]  If this is a code change, a spec testing the functionality has been added.
- [ ]  If the commit message has [changed] or [removed], there is an upgrade path above.

### DIFF
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -98,6 +98,8 @@ export default class ModalPortal extends Component {
   }
 
   componentWillUnmount() {
+    // Remove body class
+    bodyClassList.remove(this.props.bodyOpenClassName);
     this.beforeClose();
     clearTimeout(this.closeTimer);
   }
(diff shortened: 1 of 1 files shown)


diasbruno:
I'll run some tests on this. PR looks great though.

diasbruno:
Thanks, @EduardoNogueira. Released `v3.0.4`.

PR #539, opened 2017-10-23
[added] Should focus on close prop to allow focus on body instead of previous focused element

ilikepies101:
Fixes #[306].

Changes proposed:
New prop called shouldFocusOnClose, that when changed to false, puts the focus on document.body, instead of the previous element.

https://github.com/reactjs/react-modal/issues/306 touched on this issue before (still replicable in https://codesandbox.io/s/El6mDVR8g), and shouldFocusAfterRender was created. However, shouldFocusAfterRender still makes the grid jump regardless of the value. When using shouldFocusOnClose though, and setting the value to false, the grid in the example does not jump. 

Acceptance Checklist:


diasbruno:
Hi @ilikepies101 I'll review your PR tomorrow. You can submit just the actual changes and the publish task will generate all the `dists` when releasing a new version and also update changelog.

src/components/Modal.js
```diff
@@ -52,13 +52,14 @@ export default class Modal extends Component {
     onRequestClose: PropTypes.func,
     closeTimeoutMS: PropTypes.number,
     ariaHideApp: PropTypes.bool,
-    shouldFocusAfter: PropTypes.bool,
+    shouldFocusAfterRender: PropTypes.bool,
```
ilikepies101: This was misnamed because there is no variable called shouldFocusAfter

diasbruno: This is really a mismatch between Modal and ModalPortal. You can open a new PR for this one.

ilikepies101: Ok I'll do that separately. In the mean time, I'll change it back.

kloots:
@ilikepies101 @diasbruno I've got [a slightly different approach to this PR up for review](https://github.com/reactjs/react-modal/pull/540). Mind having a look?

diasbruno:
@kloots I think you can collab on @ilikepies101's branch to add this feature and repurpose yours to fix the `shouldFocusAfterRender` and those other minor issues tests and lint issues.

```js
const shouldReturnFocus = !this.props.shouldFocusAfterRender ||
                          this.props.shouldReturnFocusAfterClose;
```

This is correct from @kloots patch.

@ilikepies101 you can add @kloots as a collaborator on your fork.

ilikepies101:
Sounds good! Just added @kloots as a collaborator.

kloots:
@ilikepies101 @diasbruno, from looking at this PR it seems like it does the same as mine, except mine goes a bit further in that it won't try to restore focus if we never tried to move it into the dialog in the first place. Should we just merge mine?

diasbruno:
@ilikepies101 Thank you so much for contributing with `react-modal`. Unfortunately, #540 has supersede this one.

PR #540, opened 2017-10-23
[Fixed issue #464]

### LINKED ISSUE #464: Allow opting out of returning focus on close
### Summary:

I use react-modal to display a search box with autocompletion. When I find a suitable option and hits enter key, I'm navigated to corresponding view, and the modal should be closed. Unfortunately, button that opens the modal gets re-focused, and a click event is fired on it (because that's how enter key works with buttons), so the modal gets open again.

It helps if I add some closeTimeoutMS, but then it starts depending on how long do I hold the key. So I'd prefer just to opt out of refocusing in such cases.

### Link to example of issue:
https://codesandbox.io/s/kZmzPBLGX

kloots:
Fixes #464.

Changes proposed:
- Adds a new prop `shouldReturnFocusAfterClose` allowing explicit control over if a modal should restore focus to the element that had focus prior to its displayed.
- Does not try to restore focus if focus wasn't transferred to the modal in the first place (`shouldFocusAfterRender` was set to `false`)
- Fixes a typo in the `Modal` props so that `shouldFocusAfter` is now correctly `shouldFocusAfterRender`
- Fixes a number of line-length warnings

Acceptance Checklist:

### DIFF
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -44,6 +44,7 @@ export default class ModalPortal extends Component {
     closeTimeoutMS: PropTypes.number,
     shouldFocusAfterRender: PropTypes.bool,
     shouldCloseOnOverlayClick: PropTypes.bool,
+    shouldReturnFocusAfterClose: PropTypes.bool,
     role: PropTypes.string,
     contentLabel: PropTypes.string,
     aria: PropTypes.object,
@@ -137,8 +138,23 @@ export default class ModalPortal extends Component {
   afterClose = () => {
     // Remove body class
     bodyClassList.remove(this.props.bodyOpenClassName);
-    focusManager.returnFocus();
-    focusManager.teardownScopedFocus();
+
+    if (this.shouldReturnFocus()) {
+      focusManager.returnFocus();
+      focusManager.teardownScopedFocus();
+    }
+  };
+
+  shouldReturnFocus = () => {
+    // Don't restore focus to the element that had focus prior to
+    // the modal's display if:
+    // 1. Focus was never shifted to the modal in the first place
+    //    (shouldFocusAfterRender = false)
+    // 2. Explicit direction to not restore focus
+    return (
+      this.props.shouldFocusAfterRender ||
+      this.props.shouldReturnFocusAfterClose
+    );
   };
 
   open = () => {
@@ -147,8 +163,11 @@ export default class ModalPortal extends Component {
       clearTimeout(this.closeTimer);
       this.setState({ beforeClose: false });
     } else {
-      focusManager.setupScopedFocus(this.node);
-      focusManager.markForFocusLater();
+      if (this.shouldReturnFocus()) {
+        focusManager.setupScopedFocus(this.node);
+        focusManager.markForFocusLater();
+      }
+
       this.setState({ isOpen: true }, () => {
         this.setState({ afterOpen: true });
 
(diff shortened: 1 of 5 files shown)


diasbruno:
Hi @kloots There are 3 PRs in one. If you can stack 3 commits, one for each fix, it will show up better on the log and changelog.

diasbruno:
`[added] shouldReturnFocusAfterClose to control focus.`
`[fixed] correct property shouldFocusAfterRender.`
`[chore] refactoring tests and lint.`

Something like this.

diasbruno:
...and thank you for taking a look on this.

diasbruno:
This line also needs a condition [ModalPortal.js#L151](https://github.com/reactjs/react-modal/blob/master/src/components/ModalPortal.js#L151), otherwise the `activeElement` will be pushed on the focus stack and might cause troubles when open/close other modals (unlikely but possible).

kloots:
@diasbruno All your suggestions are done, done and done

kloots:
@diasbruno any reason this hasn't been merged yet? Asking because we need this for Slack.

diasbruno:
@kloots @ilikepies101 I'd be in favor of the first PR, just because it was the first...but it seems that the other PR will need to be rebased and also this one is more complete. So, I'll merge this one.

Thank you both for working on this issue.

diasbruno:
Released `v3.1.0`.

src/components/ModalPortal.js
```diff
@@ -137,8 +138,23 @@ export default class ModalPortal extends Component {
...
+
+  shouldReturnFocus = () => {
+    // Don't restore focus to the element that had focus prior to
+    // the modal's display if:
+    // 1. Focus was never shifted to the modal in the first place
+    //    (shouldFocusAfterRender = false)
+    // 2. Explicit direction to not restore focus
+    return (
+      this.props.shouldFocusAfterRender ||
+      this.props.shouldReturnFocusAfterClose
```
Hypnosphi: I believe this should be `&&` not `||`. Currently I have to pass both `shouldFocusAfterRender={false}` and `shouldReturnFocusAfterClose={false}` to opt out of refocusing.

diasbruno: Looking at this now, if `shouldReturnFocusAfterClose=false` would also disable the 'focus after render'. They should be able to be independent of each other.

diasbruno: NOTES:

If `shouldFocusAfterRender=true and shouldReturnFocusAfterClose=false`, we put the element in the focus manager and on close we just pop it out.

`shouldFocusAfterRender=false and shouldReturnFocusAfterClose=true` must be noop just like `shouldFocusAfterRender=false and shouldReturnFocusAfterClose=false`.

`shouldFocusAfterRender=true and shouldReturnFocusAfterClose=true` is the default behavior.

Does this make sense?

Hypnosphi: Yeah, sounds reasonable

Hypnosphi: Btw, what are the usecases for `shouldFocusAfterRender`? Looks like it just focuses a [div](https://github.com/reactjs/react-modal/blob/master/src/components/ModalPortal.js#L310-L317) without any keyboard handlers

diasbruno: To set the scope for tabbing, so it would leave the modal until it is open.

PR #564, opened 2017-11-23
Prevent focus from leaving modal by clicking on overlay.

### LINKED ISSUE #186: Pressing `esc` key won't close the modal if the `shouldCloseOnOverlayClick` is set to `false`
### Summary:

It potentially might be a bug or just a question, I have a modal that the `shouldCloseOnOverlayClick={false}`, obviously clicking on overlay will not close the modal, but pressing on an `esc` key after that will not close the modal as well. But, opening a modal and pressing the `esc` key alone without clicking on overlay will close the modal as expected.
### Steps to reproduce:
1. Creates a new `react-modal` component.
2. Set component option of `shouldCloseOnOverlayClick={false}`.
3.  Run server, open the modal, and clicking on the overlay layer.
4. Press the `esc` button.
### Expected behavior:

The modal should close.

diasbruno:
Fixes #186.

If the user doesn't allow close when clicking on the
overlay, we need to stop the event so the focus won't leave the modal
(where the key handler is defined).

Changes proposed:
- Prevent the mouse event when `shouldCloseOnOverlayClick` when set to false.

- None

Acceptance Checklist:

### DIFF
--- a/examples/basic/simple_usage/index.js
+++ b/examples/basic/simple_usage/index.js
@@ -76,6 +76,7 @@ class SimpleUsage extends Component {
           closeTimeoutMS={150}
           contentLabel="modalB"
           isOpen={currentModal == MODAL_B}
+          shouldCloseOnOverlayClick={false}
           onAfterOpen={this.handleOnAfterOpenModal}
           onRequestClose={this.toggleModal(MODAL_B)}>
           <h1 id="heading" ref={h1 => this.heading = h1}>This is the modal 2!</h1>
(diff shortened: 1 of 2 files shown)


diasbruno:
@antonhalim The issue is a bit old, but now I've found time to fix it. The problem was when clicking on the overlay, it changes the `document.activeElement` to the `<body />`. The solution prevents this change so the activeElement won't leave the modal.

diasbruno:
BTW, it still reproducible on `v3+`.

PR #570, opened 2017-11-29
Improve accessibility, fixes #359

### LINKED ISSUE #359: aria-hidden=true on body breaks Accessibility <body class="ReactModal__Body--open" aria-hidden="true">

sheerun:
- adds aria-modal="true" to modal portal
- doesn't make body a default appElement
- warns if appElement is not set in any way

Fixes #359

Acceptance Checklist:
Unnecesary rule: You can squash with button when merging
Unnecessary rule: You can edit commit message as well :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,9 +57,6 @@ to prevent assistive technologies such as screenreaders
 from reading content outside of the content of
 your modal.
 
-It's optional and if not specified it will try to use
-`document.body` as your app element.
-
 If you are doing server-side rendering, you should use
 this property.
 
--- a/package.json
+++ b/package.json
@@ -62,7 +62,8 @@
   },
   "dependencies": {
     "exenv": "^1.2.0",
-    "prop-types": "^15.5.10"
+    "prop-types": "^15.5.10",
+    "warning": "^3.0.0"
   },
   "peerDependencies": {
     "react": "^0.14.0 || ^15.0.0 || ^16",
--- a/specs/Modal.spec.js
+++ b/specs/Modal.spec.js
@@ -339,15 +339,6 @@ export default () => {
     unmountModal();
   });
 
-  it("uses document.body for aria-hidden if no appElement", () => {
-    ariaAppHider.documentNotReadyOrSSRTesting();
-    const node = document.createElement("div");
-    ReactDOM.render(<Modal isOpen />, node);
-    document.body.getAttribute("aria-hidden").should.be.eql("true");
-    ReactDOM.unmountComponentAtNode(node);
-    should(document.body.getAttribute("aria-hidden")).not.be.ok();
-  });
-
   it("raises an exception if the appElement selector does not match", () => {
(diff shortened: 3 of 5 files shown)


sheerun:
I've updated the readme as well

sheerun:
Also fixed prettier issues

sheerun:
also satisfied eslint

diasbruno:
Fantastic! I have to finish the review later, but I'll make sure we release this ASAP.

diasbruno:
Running some tests, everything looks great. The only question I have is making `setAppElement` required. 

`react-modal` has become more flexible in terms of accessibility, like a user may want to disable some a11y feature on a edge case...so, one option is to not set `aria-hidden` if `appElement` is not defined. 

We could write a log message, on dev mode, explaining why not defining `appElement` it is bad.

Any thoughts?

sheerun:
@diasbruno I didn't make setAppElement required. I just show a warning if it's not set.

I also updates PR so it doesn't show warning in production, tells how to opt-out with ariaHideApp and tells why setting this is necessary for accessibility. The message says now:

```
        "react-modal: App element is not defined.",
        "Please use `Modal.setAppElement(el)` or set `appElement={el}`.",
        "This is needed so screen reades don't see main content when modal is opened.",
        "It is not recommended, but you can opt-out by setting `ariaHideApp={false}`."

```

diasbruno:
Oh, ok...what about the second part (don't set `aria-hidden` if `appElement` is not defined)? Having `ariaHideApp={false}` is great, but it will require users to fix their code (breaking change). Any ideas?

sheerun:
@diasbruno As I said it's not a breaking change as it only shows a warning, and only in development.

diasbruno:
Ah, ok. It will be `noop` in this case.

diasbruno:
Sorry, I wasn't paying attention to this case. I was think on something more explicit like `appElement != null && ariaHideApp`, but this way looks better.

sheerun:
There is:

```
    if (ariaHideApp) {
      ariaAppHider.hide(appElement);
    }
```

but feel free to change it (this PR is editable by you). It seems artificial to me to pass ariaHideApp variable to show/hide, just to do simple `&& ariaHideApp`

diasbruno:
Thank you so much @sheerun!

diasbruno:
Release `v3.1.6`.

PR #586, opened 2017-12-19
[fixed] stop propagating ESC key event.

### LINKED ISSUE #583: onRequestClose closes all modals when using them recursively.
### Summary:
When using nested react-modals, closing with escape key closes all modals instead of the last one open.
### Steps to reproduce:
 1. click open modal button
 2. inside Modal1, click open modal 2 button
 3. When modal 2 is open click escape key for close it.

### Expected behavior:
Only modal 2 is closed and modal 1 stays open.
### Link to example of issue:
https://codesandbox.io/s/w70xm5y1y7

### additional notes
I suppose this is related to this issue, when bubbling events with Modals it gets to the parent component. This behaviour depending on the situation is quite problematic. I think it didn't happened with old version before react 16 CreatePortal but now it does. Gonna try to create another codesandbox to confirm it.

diasbruno:
Fixes #583.

Changes proposed:
- Stop event propagation when on `handleKeyDown`. This is now required when nesting modals.

- None

Acceptance Checklist:

### DIFF
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -207,7 +207,7 @@ export default class ModalPortal extends Component {
     }
 
     if (this.props.shouldCloseOnEsc && event.keyCode === ESC_KEY) {
-      event.preventDefault();
+      event.stopPropagation();
       this.requestClose(event);
     }
   };
--- a/examples/basic/app.js
+++ b/examples/basic/app.js
@@ -5,6 +5,7 @@ import SimpleUsage from './simple_usage';
 import MultipleModals from './multiple_modals';
 import Forms from './forms';
 import ReactRouter from './react-router';
+import NestedModals from './nested_modals';
 
 const appElement = document.getElementById('example');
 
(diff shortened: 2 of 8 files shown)


src/components/ModalPortal.js
```diff
@@ -207,7 +207,7 @@ export default class ModalPortal extends Component {
     }
 
     if (this.props.shouldCloseOnEsc && event.keyCode === ESC_KEY) {
-      event.preventDefault();
```
Ne3l: Does Remove prevenDefault break something? Why was there before?

diasbruno: There is no need to `preventDefault` in this case, but instead to let only modal that receives this event to process it.

Ne3l:
@diasbruno in #583 do you mention a test but isn't any test in this PR, it is already commited?

diasbruno:
@Ne3l thanks. I've forgot to include the tests.

diasbruno:
Added the missing test.

diasbruno:
Thank you so much for helping.

diasbruno:
Released `v3.1.9`.

diasbruno:
Rereleased as `v3.1.10`. 
Published `3.1.9` was not released from `master` branch.

Issue #585, opened 2017-12-18
Opening modal causes body scroll top or bottom


prztrz:
I have an issue that wehever I open the modal the body is immediatelly scrolled  to top or to bottom of the website. When the modal is closed the user is not in the place he/she opened the modal but either on the top or on the bottom of website.

I used following CSS to prevent parent element from scrolling (it allways causes the body scrolling top when modal is opened):

```css
.ReactModal__Body--open {
    overflow: hidden;
    position: fixed;
    width: 100%;
}
```

When I try to use just ```overflow:hidden``` as recommended in readme.md it does not work for me - the parent element is still scrollable.
When I try to use it without ```position: fixed``` as below: 

```css
.ReactModal__Body--open {
    overflow: hidden;
    width: 100%;
}
```
It always causes the body scroll bottom when modal is opened.

Skipping ```width: 100%``` causes the body content is shifted to the left if ```position: fixed``` is used as well.

ryansukale:
Until now I was only seeing the issue in firefox.
But as of today, this is happening in on both firefox and chrome.

lydell:
Wild guess: Could it be that you have something like:

```jsx
<a href="#" onClick={() => this.openModal()}>Open modal</a>
```

and forgot `event.preventDefault()`? (You could also use a `<button>`.)

```jsx
<a href="#" onClick={event => {event.preventDefault(); this.openModal()}}>Open modal</a>
```

prztrz:
@lydell unfortunatelly this is not the problem in my case. Clickable elements are ```<div aria-role="button">``` those should not cause automatically reload page by default, however I tried ```event.preventDefault()``` anyway and it did not work.

For now I worked the problem around engaging state and additional scrollBack() function like here:
```
//changing modal isOpen prop to true
//param scrollPos is a number returned by window.scrollY
openModal = scrollPos => { 
    this.setState({
        isModalOpen: true,
        scrollPosOnModalOpen: scrollPos
    })
}

onRequestClose = () => {
    this.setState({
        isModalOpen: false,
   }, this.scrollBack)
}

scrollBack = () => {
     window.scrollTo(0, this.state.scrollPosOnModalOpen)
}

```

Well it works but does not look very efficient.

diasbruno:
@prztrz @lydell @ryansukale Can you check if #588 can help with this issue?

prztrz:
@diasbruno thank you, this solution solved my problem.

diasbruno:
[...]

PR #588, opened 2018-01-03
[added] add class to html when modal is open

### LINKED ISSUE #60: Block scrolling events behind the modal
Is there some way to block scrolling events behind the modal while the modal is active? I have a long list of items and each one opens a modal. If you scroll while the modal is active, even if there isn't data to scroll in the modal, it scrolls the list behind the modal. This can be replicated by just adding a bunch of paragraph tags to the example project so that you page is large. What's worse is when there is data in the modal that also scrolls. The scrolling will scroll behind the modal all the way to the top or bottom and then start scrolling the data inside the modal.

Anyone know how to fix this behavior?

Thanks!

### LINKED ISSUE #251: Prevent background scrolling when modal is displayed
I couldn't find the proper solution to prevent scrolling of the backdrop when modal is displayed. I found a few but it was confusing and long text too. How is it done?

bannier:
- Fixes #60  on browsers where `overflow:hidden` on `body` is not enough
- Fixes #251 on browsers where `overflow:hidden` on `body` is not enough

Changes proposed:
- add class to `html` when modal is open

Acceptance Checklist:

This addition involves renaming some helpers from `bodyClass` to more generic terms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,14 +187,17 @@ function getParent() {
 </Modal>
 ```
 
-### Body class
+### Body and Html class
 
 When the modal is opened a `ReactModal__Body--open` class is added to the `body` tag.
-You can use this to remove scrolling on the body while the modal is open.
+and a `ReactModal__Html--open` class is added to the `html` tag.
+You can use these to remove scrolling on the body while the modal is open.
 
 ```CSS
 /* Remove scroll on the body when react-modal is open */
-.ReactModal__Body--open {
+.ReactModal__Body--open,
+.ReactModal__Html--open {
+{
     overflow: hidden;
 }
 ```
--- a/docs/README.md
+++ b/docs/README.md
@@ -68,6 +68,11 @@ import ReactModal from 'react-modal';
   */
   bodyOpenClassName="ReactModal__Body--open"
   /*
+     String className to be applied to the document.html (must be a constant string).
+     See the `Styles` section for more details.
+  */
+  htmlOpenClassName="ReactModal__Html--open"
+  /*
     Boolean indicating if the appElement should be hidden
   */
   ariaHideApp={true}
--- a/docs/styles/classes.md
+++ b/docs/styles/classes.md
@@ -8,12 +8,15 @@ class for each of those.
 You can override the default class that is added to `document.body` when the
 modal is open by defining a property `bodyOpenClassName`.
 
-It's required that `bodyOpenClassName` must be `constant string`, otherwise we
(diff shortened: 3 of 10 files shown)


prztrz:
This solves my problem from #585 completely

diasbruno:
@bannier Thanks!! I'll review your PR later. Looks good.

diasbruno:
[https://github.com/reactjs/react-modal/blob/master/examples/basic/app.css](https://github.com/reactjs/react-modal/blob/master/examples/basic/app.css) 

This is the example of css for the tests, can you include the class for the `html`, please?

bannier:
[done](https://github.com/reactjs/react-modal/pull/588/files#diff-e2a37c66472180100b1cbae4b0c17ed1R33)

diasbruno:
I don't know if we should use `<html>` by default. It seems ok, but maybe we should let the user decide when to enable this.

```js
<Modal htmlOpenClassName={string || null} ... />
```

If `htmlOpenClassName` is not defined or null, it should be `no-op` (in this case, the default value would be `null`).

Any thoughts?

diasbruno:
cc @bannier

bannier:
sorry I haven't checked my PR since last week. Sure I can add an option for that. But 2 questions still : 
- don't you think the default should be to add this class ?
- shouldn't we also add this option for class on `<body>` too ?

diasbruno:
I don't know,...It would keep the same behavior for old applications and only enable this for projects that will need this.

Hope this make sense.

bannier:
It makes sense. I'll do it.

diasbruno:
Let me know if you need any help, @bannier?

rickerd:
Uhm.. why not merging this?? I have the same issue..

diasbruno:
@bannier @rickerd I'll merge this PR and make a new PR with the requested changes. It will take a little longer to release a new version with this feature.

bannier:
Sorry @diasbruno  I had issues with adapting tests to this options and didn't took a second shot at it since. Work in progress [here](https://github.com/bannier/react-modal/pull/1/files)

diasbruno:
No problem, @bannier. Please, let me know if you need anything.

semone:
Nice job! 
What is the status on this PR? This would solve an issue I have

alex-shamshurin:
Guys, more than 1 month has gone. Would it be finally merged or not?

diasbruno:
I'm moving this PR to a new one (#614). 

Thank you so much, @bannier, for this feature.

PR #608, opened 2018-02-09
[fixed] Tab focus escapes modal on shift + tab after opening

### LINKED ISSUE #607: Tab focus escapes modal on shift + tab after opening
### Steps to reproduce:

 1. Open modal
 2. Hit `shift` + `tab` (before doing anything else)
 3. You'll see that the tab focus will move outside the modal.

### Expected behavior:
Tab focus is _always_ trapped inside the modal when it is open.

### Link to example of issue:
This can be reproduced in all the examples in this repo.

jasonblanchard:
Fixes #607.

Changes proposed:

Focusing the last tabbable element on `shift` + `tab` if the tab focus is on the modal content by checking if the shift key is active in `scopeTab` when tabbing off the modal content. Because the modal content has `tabindex="-1"`, this _should_ only happen after focus is programmatically given to the modal content on open.

Acceptance Checklist:

### DIFF
--- a/src/helpers/scopeTab.js
+++ b/src/helpers/scopeTab.js
@@ -13,9 +13,11 @@ export default function scopeTab(node, event) {
   const head = tabbable[0];
   const tail = tabbable[tabbable.length - 1];
 
-  // proceed with default browser behavior
+  // proceed with default browser behavior on tab.
+  // Focus on last element on shift + tab.
   if (node === document.activeElement) {
-    return;
+    if (!shiftKey) return;
+    target = tail;
   }
 
   var target;
--- a/specs/Modal.events.spec.js
+++ b/specs/Modal.events.spec.js
@@ -44,6 +44,21 @@ export default () => {
     document.activeElement.should.be.eql(content);
   });
 
+  it("Traps tab in the modal one shift + tab", () => {
+    const topButton = <button>top</button>;
+    const bottomButton = <button>bottom</button>;
+    const modalContent = (
+      <div>
+        {topButton}
+        {bottomButton}
+      </div>
+    );
+    const modal = renderModal({ isOpen: true }, modalContent);
+    const content = mcontent(modal);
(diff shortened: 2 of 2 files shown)


src/helpers/scopeTab.js
```diff
@@ -13,9 +13,11 @@ export default function scopeTab(node, event) {
   const head = tabbable[0];
   const tail = tabbable[tabbable.length - 1];
 
-  // proceed with default browser behavior
+  // proceed with default browser behavior on tab.
+  // Focus on last element on shift + tab.
   if (node === document.activeElement) {
-    return;
+    if (!shiftKey) return;
+    target = tail;
```
diasbruno: I think we should prepend the `node` to the list of tabbable.

`const tabbable = [node].concat(findTabbable(node));`

This way the node will always be the head.

`if (!tabbable.length) {` => `if (tabbable.length == 1) {` (above).

jasonblanchard: Would that break the concept of what `tabbable` means, though? In tabbable.js, it looks like it's looking for things that are naturally tab focusable which would [exclude the modal content `node` that has `tabindex="-1"`](https://github.com/reactjs/react-modal/blob/master/src/helpers/tabbable.js#L50).

I'm open to making that change, though. I'm not as familiar with the history of tabbable.js, here.

diasbruno: I think the question is (I'm also not sure about this) if it should cycle only on the children of the content or if the content should be allowed to gain the focus back.

If content should be focusable, the `tabIndex="-1"` can be removed (since it will be the first to receive the focus).

diasbruno: Thanks for the clarification.

diasbruno:
Thanks for having a look on this.

diasbruno:
Released `v3.1.13`.

PR #614, opened 2018-02-20
[added] add class to html when open the modal.

diasbruno:
This is a continuation and review of #588 (feature by @bannier).

Changes proposed:

- add class to `html` tag when modal is open.
- this feature will follow the same rules as `document.body` class.

Acceptance Checklist:


diasbruno:
@ianprime0509 @bannier Can you help me review this PR? I just made the `htmlOpenClassName` to follow the same rules as `bodyOpenClassName`.

examples/basic/app.css
```diff
@@ -29,3 +29,8 @@
   transform: scale(0.5) rotateX(30deg);
   transition: all 150ms ease-in;
 }
+
+.ReactModal__Body--open,
+.ReactModal__Html--open {
```
ianprime0509: It looks like the `ReactModal__Html--open` class is never being applied to the HTML element (unless it is set explicitly); in fact, this is tested on line 279 of `specs/Modal.spec.js`.  On the other hand, the `ReactModal__Body--open` class is applied to the body by default, which seems potentially inconsistent.

specs/Modal.spec.js
```diff
@@ -253,18 +255,33 @@ export default () => {
...
+    isHtmlWithReactModalOpenClass().should.not.be.ok();
+  });
+
+  it("append class to document.body if modal is open.", () => {
     renderModal({ isOpen: true });
     isBodyWithReactModalOpenClass().should.be.ok();
-    unmountModal();
   });
 
+  it("don't append class to <html /> if not defined.", () => {
```
ianprime0509: Here's the test that I mentioned above about the HTML class not being applied by default.

examples/basic/app.css
diasbruno: At least one example must use this property, I guess.

diasbruno:
Thanks for pointing out the docs. I'll update it.

diasbruno:
I think everything should be ok now. I'll leave the example to other PR.

docs/README.md
```diff
@@ -69,6 +69,11 @@ import ReactModal from 'react-modal';
      See the `Styles` section for more details.
   */
   bodyOpenClassName="ReactModal__Body--open"
+  /*
+     String className to be applied to the document.html (must be a constant string).
+     See the `Styles` section for more details.
```
diasbruno: Should explicit tell that this attribute is `null` by default.

diasbruno: Done.

specs/Modal.spec.js
ianprime0509: Yes, that makes sense, thanks for clarifying!

diasbruno: It's more of a conservative approach, but we can later change to define it by default.

diasbruno:
Thanks @ianprime0509 and @bannier for helping with this. 👍

specs/Modal.spec.js
woniesong92: Hey @diasbruno, shouldn't doc mention that you have to explicitly define `htmlOpenClassName` for it to work?

diasbruno: @woniesong92 I updated the docs to mention that this prop is `null` by default. I'll check if the published doc is not updated, but I'm sure I publish `3.3.1` and the new docs.

docs/README.md
```diff
@@ -69,6 +69,12 @@ import ReactModal from 'react-modal';
      See the `Styles` section for more details.
   */
   bodyOpenClassName="ReactModal__Body--open"
+  /*
+     String className to be applied to the document.html (must be a constant string).
+     This attribute is `null` by default.
+     See the `Styles` section for more details.
+  */
+  htmlOpenClassName="ReactModal__Html--open"
```
diasbruno: @woniesong92 Here.

docs/styles/classes.md
```diff
@@ -61,6 +61,21 @@ non-default `bodyOpenClassName`), you could use the following CSS:
 }
 ```
 
+You can define a class to be added to the html tag, using the `htmlOpenClassName`
+attribute, which can be helpeful to stop the page to scroll to the top when open
+a modal.
```
diasbruno: @woniesong92 Ah...I think I could add something here to remind that it's `null`.

jaspersorrio:
Hello I tried using this feature today but couldn't even after npm install -save react-modal@latest.

Is the code on npm updated?

diasbruno:
Hi @jaspersorrio, can you open a issue for this, please?

ElixirMike:
Trying to follow this thread. So, what exactly do I need to do in order to not have the scroll to the top of the window when openiing the modal? It's not clear what I need to add where?

nolandg:
Has the scrolling to top on open issue ever been solved? I've tried overflow, position, tab index, everything I could find in these threadsy but nothing has worked.

PR #623, opened 2018-03-02
Update README.md

daiyanze:
docs[README.md]: Add `Modal.setAppELement` description

I think it would be quite helpful to inform the essential usage since it is updated. 

Changes proposed:
- add one line to describe setting app element before using modal

Acceptance Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,9 @@ const customStyles = {
   }
 };
 
+// Make sure to bind modal to your appElement (http://reactcommunity.org/react-modal/accessibility/)
+Modal.setAppElement('#yourAppElement')
+
 class App extends React.Component {
   constructor() {
     super();


diasbruno:
Hi @daiyanze, can you please add `[chore]` to your commit message?

Also, I think we can also improve the [accessibility section](http://reactcommunity.org/react-modal/accessibility/#app-element) on the docs.

It seems that the example uses `#root` as app element, but it would cause problems. We need to update the docs saying that the modal element should not be rendered as a descendent of `app element`

cc @MaksimKniazeu to comment (if possible).

daiyanze:
Sorry about the commit issue (not forked the project yet). I will create a new pr to add `[chore]` to commit message.

By the way, I added description to tell others that `react-modal` is not a descendent of app element to new pr.

daiyanze:
Close due to new pr

PR #624, opened 2018-03-06
Update/readme.md

daiyanze:
Fixes #[issue number].

Changes proposed:
- update ./README.md for prompting `Modal.setAppElement` implementation
- update ./docs/accessibility/README.md to explain usage of `Modal.setAppElement`

Acceptance Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,9 @@ const customStyles = {
   }
 };
 
+// Make sure to bind modal to your appElement (http://reactcommunity.org/react-modal/accessibility/)
+Modal.setAppElement('#yourAppElement')
+
 class App extends React.Component {
   constructor() {
     super();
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -50,6 +50,9 @@ const customStyles = {
   }
 };
 
+// Make sure to bind modal to your appElement (http://reactcommunity.org/react-modal/accessibility/)
```
diasbruno: Perhaps this can be rephrased like: `In case you want to use aria-hidden, make sure...`.

daiyanze: Well, I think if without setting appElement, it will throw warning all the time. So I guess using aria-hidden should be a "must".

diasbruno: Hm, ok.

diasbruno:
Thank you, @daiyanze!

PR #630, opened 2018-03-23
[fixed] portal node will be reparented if necessary on props change

### LINKED ISSUE #626: unable to change the parent of modal using parentSelector props
I am unable to change the parent node of the modal even though parentSelector accepts a function. 

Take for example this code:

```
class Example extends React.Component {
  getParentNode () {
    return document.querySelector(this.props.anchorSelector);
  }

  render () {
    <ReactModal isOpen={this.props.isLoading} parentSelector={this.getParentNode.bind(this)}>
      <p>Something here</p>
    </ReactModal>
  }
}
```

so if my initial `anchorSelector` has a value of `div.initial-container` the modal opens with no problem. however if I close the modal and change the anchorSelector to `div.second-container` it does not use the second-container at all and still uses the initial one.

gnoff:
Fixes #626 .

Changes proposed:
- Stop bailing before node reparenting in componentWillReceiveProps 

Upgrade Path (for changed or removed APIs): non-breaking-change

Acceptance Checklist:

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -123,8 +123,6 @@ export default class Modal extends Component {
   componentWillReceiveProps(newProps) {
     if (!canUseDOM) return;
     const { isOpen } = newProps;
-    // Stop unnecessary renders if modal is remaining closed
-    if (!this.props.isOpen && !isOpen) return;
 
     const currentParent = getParentElement(this.props.parentSelector);
     const newParent = getParentElement(newProps.parentSelector);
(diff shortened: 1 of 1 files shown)


diasbruno:
Thank you, @gnoff!

diasbruno:
Ah, It would be great to have one test for this.

diasbruno:
cc @gnoff

diasbruno:
After the new version is released, I'll merge this one.

diasbruno:
@gnoff I've rebased to the current master and merged. The hash is e5a80d6.

Thanks!

PR #639, opened 2018-04-01
removed un-safe lifecycle methods componentWillMount and componentWillUpdate

### LINKED ISSUE #637: Remove usages of unsafe lifecycle methods
React team recently published a blogpost: https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

TL;DR: `cWM`, `cWRP`, and `cWU` methods are considered unsafe for future async rendering and will be deprecated soon. The recommended way for open source maintainers to deal with it is to use [`react-lifecycles-compat`](https://github.com/reactjs/react-lifecycles-compat) package and migrate to new `static getDerivedStateToProps` method

kaushik94:
Implemented getDerivedStateFromProps and getSnapshotBeforeUpdate lifecycle methods using react-lifecycles-compat polyfill.

Fixes #637 .

Changes proposed (need some guidance with these):
- I removed the use of `prevProps` entirely according to the suggestion [here](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html). Which leaves us the option to use State. So I had to assign `defaultProps` to initial state and start mutating from there on.
- I implemented the `getSnapshotBeforeUpdate` method which replaces(sort of) the componentWillUpdate hook, but I'm not sure if I'm updating the `this.node` correctly?
- Now that we don't have access to `prevProps`, I'm having a trouble implementing `getParentElement` function. Since the way react works, is pass the `prevState` to `getDerivedStateFromProps` via recursion(kind of) in which case the getParentElement call gets recursed throwing an error. I need a suggestion for solid implementation of this without the use of `prevProps` IMHO. Functionality wise the current implementation is working but fails the `propTypes`(obviously).

Acceptance Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "prettier": "^1.7.4",
     "react": "^16.0.0",
     "react-dom": "^16.0.0",
+    "react-lifecycles-compat": "^1.1.0",
     "react-router": "^4.2.0",
     "react-router-dom": "^4.2.2",
     "should": "^13.1.0",
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -5,6 +5,8 @@ import ModalPortal from "./ModalPortal";
 import * as ariaAppHider from "../helpers/ariaAppHider";
 import SafeHTMLElement, { canUseDOM } from "../helpers/safeHTMLElement";
 
+import polyfill from "react-lifecycles-compat";
+
 export const portalClassName = "ReactModalPortal";
 export const bodyOpenClassName = "ReactModal__Body--open";
 
@@ -14,10 +16,12 @@ const createPortal = isReact16
   : ReactDOM.unstable_renderSubtreeIntoContainer;
 
 function getParentElement(parentSelector) {
-  return parentSelector();
+  return parentSelector() || Modal.defaultProps.parentSelector;
 }
 
-export default class Modal extends Component {
+class Modal extends Component {
+  state = Object.assign({}, Modal.defaultProps);
+
   static setAppElement(element) {
     ariaAppHider.setElement(element);
   }
@@ -76,9 +80,7 @@ export default class Modal extends Component {
     shouldCloseOnEsc: true,
     shouldCloseOnOverlayClick: true,
     shouldReturnFocusAfterClose: true,
-    parentSelector() {
-      return document.body;
-    }
+    parentSelector: () => document.body
   };
 
   static defaultStyles = {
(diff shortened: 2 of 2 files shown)


src/components/Modal.js
```diff
@@ -120,30 +122,44 @@ export default class Modal extends Component {
...
+  getSnapshotBeforeUpdate(prevProps, prevState) {
+    if (!canUseDOM) {
+      return null;
+    } else {
+      this.node.className = prevState.portalClassName;
+      return prevState.portalClassName;
     }
   }
 
+  componentDidUpdate(prevProps, prevState) {
```
diasbruno: Here you use an eslint disable command for unused variables.

kaushik94:
@diasbruno fixed eslint issue, please review. Also the changes proposed in #630 can be rebased here since `componentWillRecieveProps` is unsafe and dropped in favour of `getDerivedStateFromProps`

diasbruno:
@kaushik94 tomorrow I run some tests, so we can merge this.

kaushik94:
@diasbruno awesome

diasbruno:
Tested and approved!

diasbruno:
Thanks so much @kaushik94!

src/components/Modal.js
```diff
@@ -120,30 +122,46 @@ export default class Modal extends Component {
...
 
-  componentWillUpdate(newProps) {
-    if (!canUseDOM) return;
-    if (newProps.portalClassName !== this.props.portalClassName) {
-      this.node.className = newProps.portalClassName;
+  getSnapshotBeforeUpdate(prevProps, prevState) {
+    if (!canUseDOM) {
+      return null;
+    } else {
+      this.node.className = prevState.portalClassName;
```
Hypnosphi: this method should be used for reading DOM not for mutating it

Hypnosphi: See https://reactjs.org/docs/react-component.html#getsnapshotbeforeupdate and https://twitter.com/dan_abramov/status/981712092611989509

kaushik94: @Hypnosphi I did the exact same thing you mentioned before making this PR, and yes I read the links already. But the problem I am facing is that cDU is not at all called(not with the `snapshot` parameter). Also it is not documented as to how to use the snapshot in react properly. This could most probably be a bug in `pollyfil-react-compat` library or I am doing something wrong. Please feel free to make a PR on this.

kaushik94:
@diasbruno this feels so warm. Thanks a ton again!

src/components/Modal.js
Hypnosphi: There's no usecase for snapshotting here, because you don't read any data from DOM, you just change a className. It should work OK in cDU:

```
componentDidUpdate(prevProps) {
    if (!canUseDOM) return;
    if (prevProps.portalClassName !== this.props.portalClassName) {
      this.node.className = this.props.portalClassName;
    }
  }
```

src/components/Modal.js
```diff
@@ -120,30 +122,46 @@ export default class Modal extends Component {
...
+      portalClassName: newProps.portalClassName
+    };
+
     if (newParent !== currentParent) {
       currentParent.removeChild(this.node);
       newParent.appendChild(this.node);
     }
 
+    const newState = Object.assign(prevState, enumeratedState);
     !isReact16 && this.renderPortal(newProps);
```
Hypnosphi: Looks like this should be done in cDU as well. gDSFP should be pure

package.json
```diff
@@ -52,6 +52,7 @@
     "prettier": "^1.7.4",
     "react": "^16.0.0",
     "react-dom": "^16.0.0",
+    "react-lifecycles-compat": "^1.1.0",
```
Hypnosphi: This is `devDependencies`, so user apps will break unless they have this package installed on their side. I'll add an eslint rule to prevent such things

@diasbruno you may want to hold off the release until fixing that

diasbruno: Sure, no problem. Let me know when it's ready. 

cc @kaushik94

Hypnosphi: Done as a part of #642

src/components/Modal.js
kaushik94: @Hypnosphi sorry I didn't get the time to get back to this. Yes this makes sense as done in #642 .

package.json
kaushik94: Yes my bad, thanks for fixing all this 👍

PR #645, opened 2018-04-16
Added default prop for defaultStyles property

6graNik:
Fixes #[issue number].
No issue

Changes proposed:
I added defaultProp for `"defaultStyles"` prop because in render method when classname dosn't provided we are trying to access `"defaultStyles.content"` and `"defaultStyles.overlay"` properties which will cause `"TypeError"`

None


diasbruno:
Thanks, @6graNik.

PR #665, opened 2018-05-24
[chore] Fix misleading codepen example in readme

### LINKED ISSUE #660: Readme example shouldCloseOnOverlayClick does not work
### Steps to reproduce:

 1. Go to https://codepen.io/claydiffrient/pen/woLzwo
 2.  Press Trigger modal 
 3. Press outside the box, on the overlay - it does not close the modal!

ajthinking:
From ```shouldCloseOnOverlayClick={false}``` to ```shouldCloseOnOverlayClick={true}```
Fixes #660. Closes #660.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ demonstrate various features of react-modal:
 * [Minimal example](https://codepen.io/claydiffrient/pen/KNxgav)
 * [Using setAppElement](https://codepen.io/claydiffrient/pen/ENegGJ)
 * [Using onRequestClose](https://codepen.io/claydiffrient/pen/KNjVBx)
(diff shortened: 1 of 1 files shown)


diasbruno:
Hi @ajthinking,  you can add both cases on the codepen. By default, `shouldCloseOnOverlayClick` is `true`, so I think it's really nice to also show the case where it's overriding this behavior.

ImanMh:
I don't understand what's the point of this PR? as @diasbruno said `true` is the default value for `shouldCloseOnOverlayClick` so there is no need to create a codepen for it. It has already been demonstrated in every other example of the project. What I like to see instead is a table of description of all props with their defaults and their types and also a brief description.

Issue #530, opened 2017-10-09
v3 don't respect `closeTimeoutMS` on unmount anymore


Kerumen:
### Summary:

Before the React 16 compatibility, we were able to unmount the modal with an animation. It's not possible anymore with the `v3`.

### Steps to reproduce:

 1. Add a `<Modal />` with `closeTimeoutMS={200}`
 2. Tweak the CSS to display an open and a close animation
 3. Mount/unmount the modal

### Expected behavior:

The Modal should unmount with an animation.

### Link to example of issue:

React 15 and v2: https://codesandbox.io/s/7k63p8lk81 (working)
React 16 and v3: https://codesandbox.io/s/mml6ql88py (not working)

diasbruno:
Hi @Kerumen, prefer to using `isOpen` instead of conditional render for modal for now. Need to check when modal is forcible unmounted case.

Kerumen:
Yes I know. In my case I can't really use `isOpen`.
And as it used to work, I report this as a regression. :)

diasbruno:
On `src/components/Modal.js` there is a `setTimeout`. Can you dig in a little bit (if you have time)?

Kerumen:
I tried but no luck. Event with the line commented out the modal seems to unmount instantly.
Maybe this is a change on how React handle `willUnmount`?

diasbruno:
Possibly. I'll try something this evening.

TeamChad:
Hey @diasbruno. Is there any movement on this bug? We are really keen to upgrade our project to React 16 and this is the only issue holding us up :-) We are keen to keep using this great library!

diasbruno:
@TeamChad Can you do some debugging this issue? I didn't have time yet to have look at this one (sorry), but I can give some help...One place to look at is the documentation of `createPortal`.

diasbruno:
Still didn't find a solution for this issue. We can open a discussion about "why using conditional render for a modal?". Just to get some information/use case for this. cc @TeamChad @Kerumen.

REEMERGEMEDIA:
@Kerumen If you change your rendering of the modal to what's below this will solve the issue in your "not working" codesandbox.io link. I had used conditional rendering and needed to make this change after updating React.

``` render() {
    const { modal } = this.state;
    return (
      <div>
        <button onClick={() => this.toggleModal()}>
          Open Example modal
        </button>
        <Modal
          closeTimeoutMS={200}
          isOpen={this.state.modal}
[...]

PR #672, opened 2018-06-29
[chore] Update doc to inform v3 users about close transition

hoodwink73:
Does not fix any issue. But informs the user about the discussion in #530 

Changes proposed:
-  Update to documentation.  Inform users of `v3` be aware that close transition does not work with conditional render

### DIFF
--- a/docs/styles/transitions.md
+++ b/docs/styles/transitions.md
@@ -20,11 +20,49 @@ and fade out when it is closed:
 }
 ```
 
-In order for the close transition to take effect, you will also need to pass
-the `closeTimeoutMS={150}` prop to each of your modals.
 
 The above example will apply the fade transition globally, affecting all modals
 whose `afterOpen` and `beforeClose` classes have not been set via the
 `className` prop.  To apply the transition to one modal only, you can change
 the above class names and pass an object to your modal's `className` prop as
 described in the [previous section](classes.md).
+
+In order for the close transition to take effect, you will also need to pass
+the `closeTimeoutMS={150}` prop to each of your modals.
+
+Also, if you are using `v3` which supports **React 16**, the close transition works [only if you use](https://github.com/reactjs/react-modal/issues/530#issuecomment-335208533) the `isOpen` prop to toggle the visibility of the modal.
+
+Do not conditionally render the `<Modal />`.
+
+Instead of this
+
+```javascript
+{
+  this.state.showModal ?
+  <Modal
+    closeTimeoutMS={200}
+    isOpen
+    contentLabel="modal"
+    onRequestClose={() => this.toggleModal()}
+  >
+    <h2>Add modal content here</h2>
+  </Modal>
+  : null
+}
+```
+
+*Do this*
+
+```javascript
+{
+  <Modal
+    closeTimeoutMS={200}
+    isOpen={this.state.showModal}
+    contentLabel="modal"
+    onRequestClose={() => this.toggleModal()}
+  >
+    <h2>Add modal content here</h2>
+  </Modal>
+  : null
+}
+```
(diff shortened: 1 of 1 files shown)


diasbruno:
Thanks, @hoodwink73. I'll have a look on your PR later.

docs/styles/transitions.md
```diff
@@ -20,11 +20,49 @@ and fade out when it is closed:
...
-the `closeTimeoutMS={150}` prop to each of your modals.
 
 The above example will apply the fade transition globally, affecting all modals
 whose `afterOpen` and `beforeClose` classes have not been set via the
 `className` prop.  To apply the transition to one modal only, you can change
 the above class names and pass an object to your modal's `className` prop as
 described in the [previous section](classes.md).
+
+In order for the close transition to take effect, you will also need to pass
+the `closeTimeoutMS={150}` prop to each of your modals.
```
diasbruno: Maybe this can be rephrased, something like: `The close animation is controlled via timer which you can define using `closeTimeoutMS={number}`, where the value is the time interval in milliseconds. The value must be synchronized with the css transition you have specified.

What do you think?

docs/styles/transitions.md
```diff
@@ -20,11 +20,49 @@ and fade out when it is closed:
...
+
+Also, if you are using `v3` which supports **React 16**, the close transition works [only if you use](https://github.com/reactjs/react-modal/issues/530#issuecomment-335208533) the `isOpen` prop to toggle the visibility of the modal.
+
+Do not conditionally render the `<Modal />`.
```
diasbruno: Maybe we need to elaborate here because it works for react `15`, but not in `16` (if I remember correctly).

It seems that the conditional rendering in react `16+` will destroy the rendered elements unconditionally (seems the reason why it isn't work anymore). So, we probably hold the reference for the modal DOM tree, but it's not attached to any other element (this needs to be confirmed).

docs/styles/transitions.md
```diff
@@ -20,11 +20,49 @@ and fade out when it is closed:
...
+{
+  <Modal
+    closeTimeoutMS={200}
+    isOpen={this.state.showModal}
+    contentLabel="modal"
+    onRequestClose={() => this.toggleModal()}
+  >
+    <h2>Add modal content here</h2>
+  </Modal>
+  : null
```
diasbruno: `: null` can be removed.

docs/styles/transitions.md
diasbruno: The example is really awesome. :+1:

docs/styles/transitions.md
diasbruno: Ah, also this part:

The value must be synchronized with the css transition or style you have specified (because styles can also be used in react-modal).

docs/styles/transitions.md
hoodwink73: Hey there

Does this look like a legit explanation to you regarding the limitation to control the unmounting of Portal?
facebook/react#10826

And this leads to the conditional rendering of `<Modal />` not working.

I will try to do some debugging and experiment here.

diasbruno: Great! This can be included in the doc.

hoodwink73: Awesome.

diasbruno: You can also answer #530, depending on the conclusion of your investigation and eventually close the issue. I'll continue to follow this and try to help where I can.

hoodwink73:
@diasbruno hey I did some debugging and figured out concrete reasons why conditional rendering is not working in React 16.

I will add this to the opened issue.

Thanks a lot for all help. Have a great weekend.

diasbruno:
Thank you so much. Great effort on getting this, @hoodwink73.

hoodwink73:
🙌

I also posted a lengthy comment on #530 regarding my debugging efforts. It might help you to close the issue.

diasbruno:
Yeah, that was great. That should clarify what is going on. Great job.

PR #674, opened 2018-07-27
Remove aria-modal attribute to prevent browser bugs

### LINKED ISSUE #654: Improvements for aria-modal
Based on the specification, the ARIA property [aria-modal](https://www.w3.org/TR/wai-aria-1.1/#aria-modal) must be used on the element that has the `role="dialog"` (or `alertdialog`). Instead, `react-modal` uses it on the element that contains the modal, see #359 / #570.

https://www.w3.org/TR/wai-aria-1.1/#aria-modal
> Used in Roles:  window (abstract role)
> Inherits into Roles: alertdialog dialog

See also the example in the ARIA Authoring Practices 1.1:
https://www.w3.org/TR/wai-aria-practices/#dialog_modal
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

> Notes on aria-modal and aria-hidden

> - The aria-modal property was introduced in ARIA 1.1. As a new property, screen reader users may experience varying degrees of support for it.
> - **Applying the aria-modal property to the dialog element** replaces the technique of using aria-hidden on the background for informing assistive technologies that content outside a dialog is inert.
> - In legacy dialog implementations where aria-hidden is used to make content outside a dialog inert for assistive technology users, it is important that:
>   - aria-hidden is set to true on each element containing a portion of the inert layer.
>   - The dialog element is not a descendant of any element that has aria-hidden set to true.

So, to my understanding, while we wait for browsers and assistive technologies to adapt to this new property, keep using `aria-hidden` on the content outside the modal is still recommended. However, `aria-modal` should be set on the element that _is_ the modal (i.e. the one that can receive the `role="dialog"`).

Leimi:
Fixes #654.

Changes proposed:

- remove the `aria-modal` attribute set on the modal wrapper

There are a couple problems with the aria-modal attribute:

- it is not set on the correct dom element (in theory, it should be
set on the element having the dialog role, as stated in #654)
- it is buggy in Safari. The easy way to fix the problem is to remove the attribute
(again, as suggested in #654). While it's cool on paper, the attribute is currently
not necessary, since the `aria-hidden="true"` is set correctly on the
app element.

~Upgrade Path (for changed or removed APIs):~

Acceptance Checklist:
- [x] Documentation (README.md) and examples have been updated as needed. *There were no changes needed*
- [x] If this is a code change, a spec testing the functionality has been added. *As there were no test about the presence of aria-modal before, I did not change anything*

### DIFF
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -334,7 +334,6 @@ export default class ModalPortal extends Component {
         style={{ ...overlayStyles, ...this.props.style.overlay }}
         onClick={this.handleOverlayOnClick}
         onMouseDown={this.handleOverlayOnMouseDown}
-        aria-modal="true"
       >
         <div
           ref={this.setContentRef}


diasbruno:
Thank, @Leimi. I'll merge this soon.

PR #683, opened 2018-08-27
[fixed] set default role for accessibility

### LINKED ISSUE #653: Improvements for role-dialog
Follow up to #237 

NVDA 2017.4 changed the way it handles role=dialog and now it doesn’t switch to focus mode any longer. This is mentioned in the releases notes, see https://www.nvaccess.org/files/nvda/releases/2017.4/nvda_2017.4_changes.html where it’s stated: “Web dialogs in Firefox and Chrome web browsers now automatically use browse Mode, unless inside of a web application.”. See also the related Github issue: https://github.com/nvaccess/nvda/issues/4493

To my understanding, this change is because of the definition for dialog has changed in ARIA 1.1. 

Please also see a very interesting WebAIM discussion linked on the NVDA GitHub issue, specifically the comment by Birkir Gunnarsson linked there (Feb 2017): https://webaim.org/discussion/mail_message?id=33916

> As far as I know NVDA is the only screen reader to go into application mode when focus is moved to a dialog.

So to my understanding, after the change in NVDA 2017.4, all the major screen readers are now able to announce textual content within a role=dialog without having to switch mode. This means there's no more reason to omit the role=dialog unless it's explicitly passed as a prop as it was done in #237.

Given this important chage, I'd propose to set a default `role="dialog"` but also allow developers to pass a different value, for example `alertdialog` or omit it entirely passing `null`.

wszgxa:
Fixes #653 .

Changes proposed:

- Set default value for `role`

As @afercia suggested, we should set a default value for the dialog. And omit it entirely when passing `null`.

No API changed

Acceptance Checklist:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -99,6 +99,7 @@ import ReactModal from 'react-modal';
   shouldReturnFocusAfterClose={true}
   /*
     String indicating the role of the modal, allowing the 'dialog' role to be applied if desired.
+    This attribute is `dialog` by default.
   */
   role="dialog"
   /*
--- a/specs/Modal.spec.js
+++ b/specs/Modal.spec.js
@@ -117,6 +117,19 @@ export default () => {
     contentAttribute(modal, "role").should.be.eql("dialog");
   });
 
+  // eslint-disable-next-line max-len
+  it("renders the modal content with the default aria role when not provided", () => {
+    const child = "I am a child of Modal, and he has sent me here...";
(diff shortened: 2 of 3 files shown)


diasbruno:
@wszgxa I think it's just missing some tests to omit the `role` property. Can you include a test for it, please?

wszgxa:
@diasbruno Done

diasbruno:
Good job, @wszgxa. Thank you.

PR #689, opened 2018-09-26
[fixed] check if element exists before focusing in scopeTab helper

### LINKED ISSUE #687: Desktop Safari - Error in scopeTab.js
If focus is set to an element that isn't focusable, when the user tabs the variable x in scopeTab.js will be set to -1, then when tabbable[x].focus() is called an error will occur.

DarudeGarystorm:
Fixes #687

Changes proposed:
- check if element is undefined before focusing it

Acceptance Checklist:

### DIFF
--- a/src/helpers/scopeTab.js
+++ b/src/helpers/scopeTab.js
@@ -63,8 +63,13 @@ export default function scopeTab(node, event) {
   }
 
   // If the tabbable element does not exist,
-  // let the browser control the focus
-  if (typeof tabbable[x] === "undefined") return;
+  // focus head/tail based on shiftKey
+  if (typeof tabbable[x] === "undefined") {
+    event.preventDefault();
+    target = shiftKey ? tail : head;
+    target.focus();
+    return;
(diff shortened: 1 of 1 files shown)


src/helpers/scopeTab.js
```diff
@@ -62,6 +62,10 @@ export default function scopeTab(node, event) {
     x += shiftKey ? -1 : 1;
   }
 
+  // If the tabbable element does not exist,
+  // let the browser control the focus
+  if (typeof tabbable[x] === "undefined") return;
```
diasbruno: Maybe, we actually need to focus the first element in this case, otherwise I suspect it can miss radios inputs.

DarudeGarystorm: Updated to focus head/tail based on shiftKey

diasbruno: Good catch!

diasbruno:
Thanks you, @DarudeGarystorm!

PR #707, opened 2018-10-25
[added] isNested prop

Jarch09:
Changes proposed:
- adds `isNested: bool` (optional) prop, which gives you the option of not removing the modal backdrop.

**Why?** This enables Russian doll nesting, so you can have multiple modals open on top of each other.

Acceptance Checklist:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -28,6 +28,11 @@ import ReactModal from 'react-modal';
   */
   isOpen={false}
   /*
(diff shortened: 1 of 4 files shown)


Jarch09:
Hey you have a chance to check this out?

willsmanley:
So is this going to be pulled at any point?

Does this change only remove the background effect or does it also affect modal stacking order?

PR #724, opened 2018-12-14
[added] Introduce onAfterClose callback prop

### LINKED ISSUE #708: onAfterClose prop passing
### Summary:
Hi!
ReactModal component can accept onAfterOpen prop, but I need to use something like onAfterClose but i don't see it in docs.

### Expected behavior:

Need to pass onAfterClose as prop.

fernandofleury:
Fixes #708 .

Since the `onRequestClose` won't be executed when `isOpen` changes outside of the controller scope, I think it's reasonable to give a single handler to general closing.

This would allow for some more centralised solutions like:

```js
const modalProps = {
  onAfterOpen: disableBodyScroll,
  onAfterClose: enableBodyScroll,
}
```

whereas today you have to do:

```js
const modalProps = {
  onAfterOpen: disableBodyScroll,
  onRequestClose:  enableBodyScroll
}

closeModal () {
  this.setState({ isOpen: false })
  enableBodyScroll();
}
```

Changes proposed:

- Introduce `onAfterClose` callback prop

Acceptance Checklist:

### DIFF
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -43,6 +43,7 @@ export default class ModalPortal extends Component {
     ariaHideApp: PropTypes.bool,
     appElement: PropTypes.instanceOf(SafeHTMLElement),
     onAfterOpen: PropTypes.func,
+    onAfterClose: PropTypes.func,
     onRequestClose: PropTypes.func,
     closeTimeoutMS: PropTypes.number,
     shouldFocusAfterRender: PropTypes.bool,
@@ -182,6 +183,10 @@ export default class ModalPortal extends Component {
       } else {
         focusManager.popWithoutFocus();
       }
+    }
+
+    if (this.props.onAfterClose) {
+      this.props.onAfterClose();
     }
   };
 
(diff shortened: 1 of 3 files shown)


src/components/ModalPortal.js
```diff
@@ -183,6 +184,13 @@ export default class ModalPortal extends Component {
         focusManager.popWithoutFocus();
       }
     }
+
+    if (
+      this.props.onAfterClose &&
+      typeof this.props.onAfterClose === "function"
```
diasbruno: In case it is not a function, it's preferable to let it break and let the user fix it, otherwise we might hide a bug on the user's code.

diasbruno:
Hi @fernandofleury. Thanks for implementing this. Good job!

src/components/ModalPortal.js
fernandofleury: Sure, I'll update it then!

fernandofleury:
@diasbruno I've updated as you requested

diasbruno:
Released version 3.8.1.

chris-pearce:
Is there any reason why this isn't anywhere in the docs?

http://reactcommunity.org/react-modal/?q=#installation

PR #716, opened 2018-11-16
[added] Added custom overlayElement and contentElement.

jackall3n:
Changes proposed:

- Adding `overlayElement` and `contentElement` to provide control over what DOM Element can be used for each.

**Why?** Not limited to using just `<div>`. Easier to add custom props to each element, e.g `id`, `onScroll`, etc...

Acceptance Checklist:
- [x] The commit message follows the guidelines in `CONTRIBUTING.md`.
- [x] Documentation (README.md) and examples have been updated as needed.
- [x] If this is a code change, a spec testing the functionality has been added.
- [x] If the commit message has [changed] or [removed], there is an upgrade path above.

Issue #699, opened 2018-10-11
Either document or rethink Portal usage


reverie:
### Summary:
The usage of React's Portal feature causes surprising behavior. A modal is visually and conceptually a separate interface element from the element that opens the portal. 

Example 1: If you click on a button to open a modal, should clicking on that modal bubble up to the button? No.

Example 2: If you open a modal from a form, for example to autocomplete an input to that form, should submitting the modal's form submit the outer form? No.

What makes this behavior worse than merely surprising is that it is not documented anywhere. Given that modals are separate from the elements that create them, I don't think creation of a Portal should be the default behavior of this library. If it will remain the default, it should be documented, and ideally an option to turn it off should be provided.

diasbruno:
Hi @reverie, those are good points. 

It's always possible that the documentation doesn't answer all the questions. Users always find different ways to use a libraries and/or sometimes features need clarification that is not there...  

Look interesting this idea of change how the modal is mounted on the DOM. If you can elaborate some ideas.

reverie:
Yes, I wonder whether it’s possible to avoid using Portal at all—it might not be. Some of my issues might be with Portal itself, like if it propagates events in ways that are surprising relative to the browser’s natural propagation rules.

diasbruno:
@reverie yes, I agree it can be confuse. This should be the way how synthetic events propagates (if I'm remembering correctly). If you can make a small example, I'll try to have a look and we can discuss what can be clarified in the docs and/or if there are better ways to handle this case.

jcrben:
I think this is the upstream issue for the confusing bubbling: https://github.com/facebook/react/issues/11387

Seems like a bit of a dealbreaker, altho there's a proposed workaround or two such as sebmarkbage in https://github.com/facebook/react/issues/11387#issuecomment-366142349 involves pulling the modal up higher in the tree - which doesn't seem so great as the Portal might be way below the button and moving them around breaks cohesion.

bsansouci:
After reading https://github.com/facebook/react/issues/11387 it seems to me like I cannot use react-modal since the event propagation will break my app. I don't want to register events handlers for every kind of event and call stopPropagation because it's clear that that will have side effects ([for example](https://github.com/facebook/react/issues/11387#issuecomment-366113321)).
One way to mitigate this issue is to use `react-modal-hook` and put the `ModalProvider` component as high as possible in the component tree, but that still uses portals which seem to have complex behavior that is potentially error prone. 

[...]

PR #734, opened 2019-02-01
[added] stopPropagationOnClickFor option

jcrben:
Changes proposed:
As noted at https://github.com/reactjs/react-modal/issues/699, the propagation of the React Portal is surprising. This is a workaround - in particular I couldn't figure out a way to stopPropagation for the overlay - I might've missed a trick.

I actually prefer https://github.com/reactjs/react-modal/pull/716 as it's more flexible but this addressed my specific use case. If we can get https://github.com/reactjs/react-modal/pull/716 in than I can close this PR and use a custom overlay instead.

Upgrade Path (for changed or removed APIs): N/A

Acceptance Checklist:
- [] If the commit message has [changed] or [removed], there is an upgrade path above.


jcrben:
OK, I ended up just wrapping the ReactModal with an unnecessary div with a click handler that stops propagation. So I guess I don't need this, altho this is not an ideal solution.

PR #736, opened 2019-02-07
Prevent scrolling in Chrome after closing dialog

### LINKED ISSUE #735: Chrome: Window scrolls when returning focus to an element after modal closes
### Summary:

In chrome, when the modal closes and focus is restored to an element that is scrolled to a non-default position, the window will scroll. In certain settings, such as an editor component like SlateJS or DraftJS, this jump is disorienting to users. 

### Steps to reproduce:

 1. Have focus on an element scrolled near the top of a window
 2. Open the dialog
 3. Close the dialog
 4. Observe that window scrolls to focus on the original element

### Expected behavior:

When focus is restored to an element, the window does not scroll

### Link to example of issue:

Admittedly having some trouble reproducing this in a fiddle or code sandbox. Sorry :(

### Additional notes:

Requires one line change at https://github.com/reactjs/react-modal/blob/master/src/helpers/focusManager.js#L42 to pass `{ preventScroll: true }` to `focus()`

CameronAckermanSEL:
Prevent scrolling in Chrome when returning focus to the previous active element when closing the dialog

Fixes #735 

Changes proposed:

- Add the `{ preventScroll: true }` option object to the focus call.

N/A

Acceptance Checklist:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -103,6 +103,11 @@ import ReactModal from 'react-modal';
   */
   shouldReturnFocusAfterClose={true}
   /*
+    Boolean indicating if the modal should use the preventScroll flag when restoring focus
+    to the element that had focus prior to its display.
+  */
+  preventScroll={false}
+  /*
     String indicating the role of the modal, allowing the 'dialog' role to be applied if desired.
     This attribute is `dialog` by default.
   */
--- a/examples/basic/nested_modals/index.js
+++ b/examples/basic/nested_modals/index.js
@@ -34,7 +34,7 @@ class Item extends Component {
                  labelledby: "item_title",
                  describedby: "item_info"
                }}>
-          <h1 id="item_title">Item: {itemNumber}</h1>
+          <h1 id="item_title">Item: {itemNumber + 1}</h1>
           <div id="item_info">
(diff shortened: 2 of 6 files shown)


diasbruno:
This looks good! I'll need to check the compatibility with older browsers.


Here is a place to start looking at:

[HTMLElement/focus](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus)

CameronAckermanSEL:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#Browser_compatibility

I've tested this in firefox and chrome. Most browsers don't support this flag and it's thereby ignored. Fortunately, this defect only happens in Chrome and firefox is largely unimpacted.

I unfortunately dont have a means of trying this in IE or edge.

diasbruno:
Maybe we could let the user set this flag, this way we don't change the behavior of existing applications.

```
<Modal {...} preventScroll />
```

The default value could be `false`. What do you think?

diasbruno:
@CameronAckermanSEL Sorry for the delay. I was hopping to get the next version to merge this, but I still trying to find a way to improve how the animation is handled.

I'll merge this and release a new version soon.

Thanks for the patch.  +1

mikkelrd:
@diasbruno Thanks for looking after this. Have there been any updates?

mikkelrd:
I have to apologize. Once I tested, I realized that #735 describes a slightly different problem than I was experiencing. And I believe the problem I'm seeing is due to something buggy in my feature. 🤦‍♂ Thanks again for responding!

That being said, this change seemed to me to be working as described. 👍

diasbruno:
@mikkelrd Awesome. I'll try to find sometime to merge this.

Thomas-Negrault:
I would love this feature for my use case:

I have a column of pictures. When I click one, I open a preview in a modal (like Facebook does). I have next/previous arrows in order to navigate without leaving the modal. I implemented an autoscroll to follow the navigation (if I click 10 times on the next button, the thumbnail column scroll to focus the 10th element).

But right know when I close the modal, the column scroll is focused on where I was when I first open the preview... I will try to implement a solution on my own but this flag to prevent the scroll would be very useful !

tt-tisraeli:
@diasbruno Hi! We have the same issue on our produce, do you have any ETA for merging this one?

CameronAckermanSEL:
Hey! This issue is fixed for us on our own fork, so I'm going to step away from this PR. Please feel free to resolve the merge conflicts and merge it in, or use it as a basis for a new PR later.

diasbruno:
This PR was rebased and merged manually. (https://github.com/reactjs/react-modal/commit/ff0a7f5b69ac6dd68e79e2c0968ccae6e1bebdec)

@CameronAckermanSEL thanks so much for taking the time to make this MR and sorry for taking so long to merge it.

PR #750, opened 2019-04-16
Hot reload removing portal node on storybook.

bertho-zero:
In a storybook the hot reload removes the parent before the .removeChild, which makes this mistake once in two:
`Can not read property 'removeChild' of null`

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -174,7 +174,7 @@ class Modal extends Component {
   removePortal = () => {
     !isReact16 && ReactDOM.unmountComponentAtNode(this.node);
     const parent = getParentElement(this.props.parentSelector);
-    parent.removeChild(this.node);
+    parent && parent.removeChild(this.node);
   };
 
   portalRef = ref => {


bertho-zero:
Probably related to https://github.com/facebook/react/issues/14811

diasbruno:
Hi @bertho-zero, not sure if it is a good idea. Maybe this can hide bugs when removing the portal node.

Perhaps, we can try to make a dead branch elimination:

```js
if (process.env.NODE_ENV === 'development') {
   (parent && parent.removeChild(this.node)) ||
       console.warn("Portal node has gone.");
} else {
   parent.removeChild(this.node);
}
```

What do you think? (Just trying to be safe.)

diasbruno:
Also, if you can provide more information about the setup, it might help...and thanks for reporting this issue.

bertho-zero:
This problem is not related to hot reload, I have the same problem in production as soon as the component is unmount.

I have a component that does not use `document.body` for the portal but uses a ref, like this:
```js
class GnaGna extends Component {
  getModalParent = () => this.schedulerRef.current;

  render() {
    return (
      <>
        <div ref={this.schedulerRef} />

        <ReactModal
          parentSelector={this.getModalParent}
        />
      </>
    );
  }
}
```

bertho-zero:
If I put the div before I have an error with .appendChild, if I put it after the parent disappears before calling .removeChild.

bertho-zero:
I found a workaround:
```js
class GnaGna extends Component {
  schedulerRef = React.createRef();

  modalRef = React.createRef();

  componentWillUnmount() {
    this.modalRef.current.node = null;
  }

  getModalParent = () => this.schedulerRef.current;

  render() {
    return (
      <>
        <div ref={this.schedulerRef} />

        <ReactModal
          ref={this.modalRef}
          parentSelector={this.getModalParent}
        />
      </>
    );
  }
}
```

diasbruno:
This patch was superseded by #778.

PR #774, opened 2019-10-03
[fixed] some untabbable elements are being returned from findTabbable

### LINKED ISSUE #732: Focus escapes modal when first or last element has style `display: none`
### Summary:
The 3.8.1 release of `react-modal` allows the focus to escape when the first or last tabbable element has style `display: none`

### Steps to reproduce:

 1. Open a `react-modal` with a few tabbable elements where the first or last tabbable element style set to `display: none`.
 2. Press tab or shift-tab a few times.
 3. When it gets to the first/last tabbable item, the tab or shift-tab event moves the focus outside the modal.

### Expected behavior:
The tab and shift-tab events never move the focus outside the modal

### Link to example of issue:
https://codesandbox.io/s/4x2nlqq599

flurmbo:
Fixes #732 

Changes proposed:

The bug in the example stems from the `hidesContents` function in `tabbable.js` returning false on the button with the `display: none` inline style. It is returning false because the element has the computed style `overflow: visible` (which is default value). 

The problem with this is that having `overflow: visible` doesn't mean there is any content that is  overflowing. 

I think there are a couple possible ways to fix this. One would be to add an additional check for the `display: none` style if `zeroSize === true`, which would directly solve the bug in the example.

I'm not sure that will catch every possibility, so instead I check the elements `scrollWidth` and `scrollHeight` to determine whether or not there is any overflowing content.

Acceptance Checklist:

### DIFF
--- a/specs/Modal.helpers.spec.js
+++ b/specs/Modal.helpers.spec.js
@@ -105,6 +105,15 @@ export default () => {
         elem.appendChild(input);
         tabbable(elem).should.containEql(input);
       });
+
+      it("excludes elements with overflow == visible if there is no visible content", () => {
+        const button = document.createElement("button");
+        button.innerHTML = "You can't see me!";
+        button.style.display = "none";
+        button.style.overflow = "visible";
+        elem.appendChild(button);
+        tabbable(elem).should.not.containEql(button);
+      });
     });
   });
 };
--- a/specs/Modal.helpers.spec.js
+++ b/specs/Modal.helpers.spec.js
@@ -99,9 +99,9 @@ export default () => {
 
       it("includes inputs visible because of overflow == visible", () => {
         const input = document.createElement("input");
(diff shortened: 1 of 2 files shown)


diasbruno:
@flurmbo This is going to be a hard one, but can you try to add a test case for this?

flurmbo:
Hi @diasbruno, sure, should the test just test the tabbable helper function or should it recreate the modal in the example and test that focus stays within the modal?

diasbruno:
A test case for tabbable would be great.

flurmbo:
Hi @diasbruno, I added a test case for the tabbable helper. It also looks like the previous test was editing the wrong DOM elements style (`elem` instead of `input`) so I fixed that in a separate commit.

diasbruno:
Thanks, @flurmbo. I'll merge this and make a release tomorrow.

PR #777, opened 2019-10-08
[Fixed] Screen readers cannot access components without ariaHideApp

### LINKED ISSUE #776: Request changing ariaHideApp default to false
### Summary:
 1. When placing modals without ariaHideApp={false} property, screen-reader users cannot access the modal at all because the default value is true and it adds aria-hidden DOM property to the component. 

### Expected behavior:
Modal dialogs should not have aria-hidden by default so that they are visible by screen-readers.

### Link to example of issue:
The minimum example in readme.md can reproduce this issue.

### Additional notes:
I am a blind web UX designer.
Since this component is used in many services, we have thousands of inaccessible modal dialogs at the moment.
I tried to patch it myself, but for some reason, I wasn't able to pass some of the tests. 3 tests fail even when I run the test at the unchanged master branch, so I suspect that this is happening only for my environment.
All I want to be changed is just one line. true to false. That will help us a lot.

yncat:
Fixes #776 
I updated the default props spec, but I am not sure if it is enough.

Changes proposed:

- Changing the default props "ariaHideApp" to false.

none

Acceptance Checklist:

### DIFF
--- a/specs/Modal.spec.js
+++ b/specs/Modal.spec.js
@@ -47,7 +47,7 @@ export default () => {
     const modal = ReactDOM.render(<Modal />, node);
     const props = modal.props;
     props.isOpen.should.not.be.ok();
-    props.ariaHideApp.should.be.ok();
+    props.ariaHideApp.should.not.be.ok();
     props.closeTimeoutMS.should.be.eql(0);
     props.shouldFocusAfterRender.should.be.ok();
     props.shouldCloseOnOverlayClick.should.be.ok();
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -76,7 +76,7 @@ class Modal extends Component {
     portalClassName,
     bodyOpenClassName,
     role: "dialog",
-    ariaHideApp: true,
+    ariaHideApp: false,
     closeTimeoutMS: 0,
     shouldFocusAfterRender: true,
     shouldCloseOnEsc: true,


jjwallace:
This message is found in my build:
`Warning: react-modal: App element is not defined. Please use `Modal.setAppElement(el)` or set `appElement={el}`. This is needed so screen readers don't see main content when modal is opened. It is not recommended, but you can opt-out by setting `ariaHideApp={false}`. `

diasbruno:
@yncat @jjwallace It's trick to get the `aria-hidden` correct...

According to the specification "`aria-hidden` indicates that the element and all of its descendants are not visible". Because of this, modals must be attached to a dom node different from the one that will receive the attribute.  

We have `Modal.setAppElement(element)`. This method needs to be called with the dom element that is going to receive the aria-hidden attribute.

Here is one possible solution:

HTML Code

```html
<div id="root">
  <div id="app"></div>
  <div id="modals"></div>
</div>
```

Javascript Code

```js
Modal.setAppElement("#modals");
```

Hope this solution can help.

yncat:
@jjwallace @diasbruno
I might have been misunderstanding its purpose and how the component handled aria-hidden exactly.
(1) We don't want screen readers to go out of the modal while it is open.
(2) Because of (1), we need to hide the main (before opening the modal) content.
(3) However, the component itself cannot determine which element to hide without any pointers.
(4) We have setAppElement, and we must point what to hide using that method.
Is my understanding correct? If it is, I'm going to close this PR because I took a totally wrong approach. I'm really sorry about that.

diasbruno:
You are correct, @yncat.

diasbruno:
@yncat Let me know if the solution I've posted helps with this issue, or, if you need assistance with this. Probably, there is something we can improve on our documentation about this.

yncat:
@diasbruno That works. Problem was that the service I'm fixing accessibility was using the component in an incorrect way (wrong usage of setAppElement). I am going to close this PR and the issue I had posted. Thank you very much!

PR #778, opened 2019-10-11
[fixed] Call parent.removeChild only if parent exists

### LINKED ISSUE #769: Failed removeChild function for parent when parentSelector is changed
Hi guys
In our project when we have player in fullscreen mode we want to show some modals on user actions. For showing modal in fullscreen I use `parentSelector` to change parent from body to player component. But when user leaves player in fullscreen mode with opened modal I get some error 
```
Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
```
because parent doesn't exist anymore (player) and removeChild function fails

Can be it fixed somehow? Thank you!

husseyexplores:
Fixes #769 

Should we also check before calling removeChild here (In `componentDidUpdate()`? 
```
const { prevParent, nextParent } = snapshot;
if (nextParent !== prevParent) {
  prevParent.removeChild(this.node);
  nextParent.appendChild(this.node);
}
```

I'm fairly new to contributing to open-sourced projects. Please let me know if I made any mistakes. Thank you!

---

Acceptance Checklist:

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -174,7 +174,14 @@ class Modal extends Component {
   removePortal = () => {
     !isReact16 && ReactDOM.unmountComponentAtNode(this.node);
     const parent = getParentElement(this.props.parentSelector);
-    parent && parent.removeChild(this.node);
+    if (parent) {
+      parent.removeChild(this.node);
+    } else {
+      /* eslint-disable-next-line no-console  */
+      console.warn(
+        "Cannot remove modal from the parent element as it no longer exists in the DOM"
+      );
+    }
   };
 
   portalRef = ref => {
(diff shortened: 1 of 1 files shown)


diasbruno:
Hi @husseyexplores, Thanks for the PR. As mentioned in the issue, You can add a warning telling that the element has gone.

husseyexplores:
Thanks for the follow-up!
By warning, you mean `console.warn`, right?

diasbruno:
Yep. 👍

husseyexplores:
Added the warning message. Let me know if it needs to be rephrased.

diasbruno:
Here is an example of a message we are already using [https://github.com/reactjs/react-modal/blob/master/src/components/ModalPortal.js#L93](https://github.com/reactjs/react-modal/blob/master/src/components/ModalPortal.js#L93).

This should be the last thing.

diasbruno:
Thanks, @husseyexplores. I'll review it.

PR #780, opened 2019-10-18
[feat]a better way to judge React version

cappuccino6:
Fixes #[issue number].

Changes proposed:

- it's a better way to judge React version, no longer dependent on the「createPortal」 api

Acceptance Checklist:


nagayev:
@Fe-Icy 
But "16.0">"15.9.9", because char code of 6 is greater than 5.
You can try this sample in our console.
So you can simplify your function and make it faster.

diasbruno:
@cappuccino6 This is more that what is needed in this case (maybe the name of this variable is misleading). The idea was to detect if we have available the new api `createPortal` available.

Issue #787, opened 2019-11-13
Accessibility: focus trap improvements to avoid accessing underlying page


hollowaysmith:
### Summary:

While doing an accessibility audit of a site an auditor was able to escape the modal's focus trap.

This does seem like an accessibility issue because,

1. For sighted users with screen readers the background page (below the modal) can be navigated and yet the modal itself obscures the background, and the easiest way to find the modal again is to use a mouse which they may not have.
2. For screen reader users with sight impairments it can be hard to find the modal again.

### Steps to reproduce:

 1. https://cdpn.io/holloway/fullpage/eYYPppW
 2. Open modal
 3. `CTRL-L` to move focus to address bar
 4. `<tab>` back into page and you can see the first button under the modal receives focus. This button will turn green when it receives focus.

Also  **`<SHIFT>`**+`<tab>` to navigate back into page from the bottom and the same issue happens (although that codepen doesn't show it).

### Expected behavior:

`<tab>`ing back into the page moves focus to the modal.

### Potential Fix

While a modal is open we could temporarily insert some focusable elements as the first child and last child of `<body>` that immediately move focus to the modal content. I guess these could be just `<div tabindex="0"></div>` elements.

I'm not sure how this would work with nested modals. Perhaps we'd restore focus to the most recently opened modal.

Do you want a PR?

If so, any advice on how to implement this?

I think this needs to be plain JS (not React) to ensure these `<div tabindex="0" />` elements are the first and last in `<body>`, but obviously we'd use lifecycle methods to clean up if React Modal was unmounted.

holloway:
Duplicate of #742

diasbruno:
I'm going to close this one in favor of #742.

Thanks, @hollowaysmith @holloway.

PR #791, opened 2019-12-06
[fixed] Focus trap when reentering document (#742)

### LINKED ISSUE #742: Modal doesn't trap focus when re-entering document
### Summary:
react-modal doesn't trap focus when re-entering a document that contains an open modal.

### Steps to reproduce:

 1. Open a react-modal modal
 2. Move focus outside the document, for example by clicking the mouse in your browser's address bar or pressing ctrl+l
 3. Hit tab key a number of times until focus moves back into the document

### Expected behavior:
Focus is trapped inside the still-open modal when focus moves back into the document.

### Actual behavior:
Focus moves through the document behind the modal.

### Link to example of issue:
https://mgwy6v6e.codesandbox.io/

### Additional notes:
Compare this behaviour to https://davidtheclark.github.io/react-aria-modal/demo/ which doesn't have this issue.

holloway:
Fixes #742

Changes proposed:
- Ensure modal focus trap when reentering document. This restores modal focus when navigating into the page by keyboard. See discussion on #742 and #787

- N/A

Acceptance Checklist:
- ~[ ] Documentation (README.md) and examples have been updated as needed.~ *N/A I think*
- ~[ ] If the commit message has [changed] or [removed], there is an upgrade path above.~ *N/A I think*

### DIFF
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -77,6 +77,15 @@ export default class ModalPortal extends Component {
     if (this.props.isOpen) {
       this.open();
     }
+
+    // Body focus trap see Issue #742
+    this.bodyFocusTrapBefore = document.createElement("div");
+    this.bodyFocusTrapBefore.style.position = "absolute";
+    this.bodyFocusTrapBefore.style.opacity = "0";
+    this.bodyFocusTrapBefore.setAttribute("tabindex", "0");
+    this.bodyFocusTrapAfter = this.bodyFocusTrapBefore.cloneNode();
+    this.bodyFocusTrapBefore.addEventListener("focus", this.focusContent);
+    this.bodyFocusTrapAfter.addEventListener("focus", this.focusContent);
   }
 
   componentDidUpdate(prevProps, prevState) {
@@ -151,6 +160,16 @@ export default class ModalPortal extends Component {
       ariaHiddenInstances += 1;
       ariaAppHider.hide(appElement);
     }
+
+    if (document.body.firstChild !== this.bodyFocusTrapBefore) {
+      document.body.insertBefore(
+        this.bodyFocusTrapBefore,
+        document.body.firstChild
+      );
+    }
+    if (document.body.lastChild !== this.bodyFocusTrapAfter) {
+      document.body.appendChild(this.bodyFocusTrapAfter);
+    }
   }
 
   afterClose = () => {
@@ -190,6 +209,13 @@ export default class ModalPortal extends Component {
 
     if (this.props.onAfterClose) {
       this.props.onAfterClose();
+    }
+
+    if (this.bodyFocusTrapBefore.parentElement === document.body) {
+      document.body.removeChild(this.bodyFocusTrapBefore);
+    }
+    if (this.bodyFocusTrapAfter.parentElement === document.body) {
+      document.body.removeChild(this.bodyFocusTrapAfter);
     }
   };
 
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -73,13 +73,15 @@ export default class ModalPortal extends Component {
     this.moveFromContentToOverlay = null;
 
     // Body focus trap see Issue #742
-    this.bodyFocusTrapBefore = document.createElement("div");
-    this.bodyFocusTrapBefore.style.position = "absolute";
-    this.bodyFocusTrapBefore.style.opacity = "0";
-    this.bodyFocusTrapBefore.setAttribute("tabindex", "0");
-    this.bodyFocusTrapAfter = this.bodyFocusTrapBefore.cloneNode();
-    this.bodyFocusTrapBefore.addEventListener("focus", this.focusContent);
-    this.bodyFocusTrapAfter.addEventListener("focus", this.focusContent);
+    if (typeof window !== "undefined") {
+      this.bodyFocusTrapBefore = document.createElement("div");
+      this.bodyFocusTrapBefore.style.position = "absolute";
+      this.bodyFocusTrapBefore.style.opacity = "0";
+      this.bodyFocusTrapBefore.setAttribute("tabindex", "0");
+      this.bodyFocusTrapAfter = this.bodyFocusTrapBefore.cloneNode();
+      this.bodyFocusTrapBefore.addEventListener("focus", this.focusContent);
+      this.bodyFocusTrapAfter.addEventListener("focus", this.focusContent);
+    }
   }
 
   componentDidMount() {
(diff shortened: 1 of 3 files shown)


diasbruno:
Legit. @holloway Have you found what broke the tests?

diasbruno:
@holloway Ping me if you need anything...not sure if I'm receiving the notifications from this repo or if I'm skipping without looking... :)

holloway:
@diasbruno Yeah the test bug was caused by making those trap elements in `componentDidMount` rather than the constructor (or, presumably, willMount).

So that's passing now.

src/components/ModalPortal.js
```diff
@@ -71,6 +71,15 @@ export default class ModalPortal extends Component {
...
     this.moveFromContentToOverlay = null;
+
+    // Body focus trap see Issue #742
+    this.bodyFocusTrapBefore = document.createElement("div");
+    this.bodyFocusTrapBefore.style.position = "absolute";
+    this.bodyFocusTrapBefore.style.opacity = "0";
+    this.bodyFocusTrapBefore.setAttribute("tabindex", "0");
+    this.bodyFocusTrapAfter = this.bodyFocusTrapBefore.cloneNode();
+    this.bodyFocusTrapBefore.addEventListener("focus", this.focusContent);
+    this.bodyFocusTrapAfter.addEventListener("focus", this.focusContent);
```
diasbruno: This will create the trap elements for each modal created. 

Perhaps you can move this code to a new file and create these elements
in a singleton fashion.

Another question, though, is what would happen if there are nested modals opened?

diasbruno: Maybe we need to track the lastest modal open (?)


NOTES: 
Probably we need to rewrite the way we manage nested modals.

holloway: @diasbruno Yeah nested modals was why I was thinking it might be easier to just make a set of these per-modal. They 'wrap' the `<body>` nodes, and each subsequent modal adds another wrapping layer that's associated with only a single modal so giving focus via `this.bodyFocusTrapBefore.addEventListener("focus", this.focusContent);` is easy.

Happy to change it to a singleton approach though (that would make fewer `<div/>`s at the expense of a bit more complexity)... you mentioned a new file so I guess in that file I'd have a array of ModalPortal instances... pushing new ones on in `beforeOpen` and popping off in `afterClose`, and calling `this.focusContent` on the top modal when these trap elements were focused. Is that what you thinking?

diasbruno: > I'd have a array of ModalPortal instances... pushing new ones on in beforeOpen and popping off in afterClose, and calling this.focusContent on the top modal when these trap elements were focused.

It seems the right way to go.

holloway: @diasbruno is this what you were thinking, or is this too generic?

src/helpers/bodyTrap.js
```diff
@@ -0,0 +1,48 @@
+import portalOpenInstances from "./portalOpenInstances";
+// Body focus trap see Issue #742
+
+let before, after, instances;
```
diasbruno: Ýou can initialize `instances` with an empty array.

src/helpers/bodyTrap.js
```diff
@@ -0,0 +1,48 @@
...
+
+function focusContent() {
+  if (!instances || instances.length === 0) {
+    // eslint-disable-next-line no-console
+    console.warn(`React-Modal: Open instances > 0 expected`);
+    return;
+  }
+  instances[instances.length - 1].focusContent();
```
diasbruno: Maybe we can let this be no-op if there are no modals opened, or, maybe warn only in develoment mode. What do you think? 
 
```js
if (instances.lenght > 0) {
  instances[instances.length - 1].focusContent();
}
```

src/helpers/portalOpenInstances.js
```diff
@@ -0,0 +1,51 @@
+// Tracks portals that are open and emits events to subscribers
```
diasbruno: This looks really great. :+1:

src/components/ModalPortal.js
diasbruno: It's great.

src/helpers/portalOpenInstances.js
holloway: Thanks :smile:

src/helpers/bodyTrap.js
holloway: @diasbruno I agree, and as per the pattern in ModalPortal.js I've made all consoles only print in dev mode.

diasbruno:
Thanks, @holloway. I merged your PR and will run a few tests on it. 
Don't let me forget to make a release. :)

diasbruno:
@holloway When I'm move the focus back to the page view - testing on firefox -, I'm seen that I need "a extra tab" because `<html>` is getting the focus. Is this the correct behavior or a difference between browsers?

holloway:
@diasbruno hmm... I can't reproduce that behaviour in Firefox v71 (Linux), and I haven't heard of Firefox overriding JS-controlled focus in that way. What are the exact reproduction steps? For me, in Firefox and in Chrome, it restores focus to the button that it opened from. This is a change in behaviour compared to this ReactCommunity Codepen examples? 

fyi I have tested focus management by running this to log the current element every second,

```
setInterval(function(){ console.log(document.activeElement); }, 1000);
```

And then interacting with the page to see how focus is moved around, and that part (when the page gets focus back) seems to behave as it was before this PR.

diasbruno:
That look correct. Maybe when I click on the browsers console and type `document.activeElement`.

Steps
- Open modal
- Click on the browser's address
- Tab until it reaches the page view
- Go to browser's console and type `document.activeElement` (maybe this is why I'm getting the `<html>`)

diasbruno:
Firefox 71 here.

holloway:
Ah then yes that might move focus to `<html>`. The trap elements are in `<body>` so if the focus is on `<html>`, before the trap, then they haven't walked into the focus trap yet.

There's a legitimate question about whether focusing `<html>` _should_ move focus to the modal. I'm not sure of the answer to that, but I'll ask around with some accessibility experts that I know. If so, that would need a different approach than these trap elements, however in my testing it's basically harmless to allow `<html>` to be focused as the other content under the modal is still bypassed.

I guess the idea of using a `setInterval` was to poll the activeElement without needing to interact with the console which might change the activeElement and trigger a window `blur` event and react-modal changes (ie, I see focusManager.js responds to window blur).

diasbruno:
LGTM. When we get more info about this, we can do something about this.

holloway:
@diasbruno fyi I'm told that the extra tab in Firefox is expected behaviour that we wouldn't want to interfere with when a modal is open. FF users are in the habit of hitting tab twice, so this PRs' approach is still appropriate.

(@jasonkiss)

holloway:
@diasbruno 

> Don't let me forget to make a release. :)

bump! :smile:

diasbruno:
:man_facepalming:

Thanks, @holloway.
I'll make a release today.

diasbruno:
Release `3.11.2`.

holloway:
No probs, thanks @diasbruno !

PR #805, opened 2020-04-30
[fixed] - Ensure focus does not scroll the modal

xaviergonz:
In IOS we found that the focus of an element would scroll the page so such element is brought into view.

In our case we have a grid of several views (iframes) and whenever a modal would appear on one of those views then the whole app would scroll to show the cell were the modal appeared, which is undesireable.

This PR fixes that.


diasbruno:
Thanks, @xaviergonz. 
I'll merge this later. Ping me if I forget. :)

PR #825, opened 2020-08-19
[fixed] lazily access ReactDOM.createPortal

### LINKED ISSUE #824: Bug: React Modal is using ReactDOM on the server
`Modal.js` is accessing a property of `ReactDOM` when it is first run, regardless of whether `renderPortal` is called or even whether `Modal` is instantiated.

https://github.com/reactjs/react-modal/blob/cec8833cccb3c2b2580cabcdf8fb244e0a4f2b34/src/components/Modal.js#L13

On serverside rendered apps, this code runs in NodeJS, where `ReactDOM` should not be used. If `ReactDOM` is not available, "Cannot read property createPortal of undefined" will be thrown.

To fix this, `isReact16` could be changed from a boolean to a function that returns a boolean so that the `ReactDOM.createPortal` check only runs when needed. Or the approach in  https://github.com/reactjs/react-modal/pull/780 would work too.

keeganstreet:
Fixes #824

Changes proposed:

- Access `ReactDOM.createPortal` via a function instead of directly, so it is only accessed when required, not when the script is compiled. Stops errors being thrown when bundled into script that is run in an environment without access to ReactDOM.

N/A

Acceptance Checklist:
- [ ] (N/A) Documentation (README.md) and examples have been updated as needed.
- [ ] (N/A) If this is a code change, a spec testing the functionality has been added.
- [ ] (N/A) If the commit message has [changed] or [removed], there is an upgrade path above.

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -10,7 +10,7 @@ import { polyfill } from "react-lifecycles-compat";
 export const portalClassName = "ReactModalPortal";
 export const bodyOpenClassName = "ReactModal__Body--open";
 
-const isReact16 = ReactDOM.createPortal !== undefined;
+const isReact16 = canUseDOM && ReactDOM.createPortal !== undefined;
 
 const getCreatePortal = () =>
   isReact16


diasbruno:
Maybe something like this will work: `const isReact16 = canUseDOM && ReactDOM.createPortal !== undefined;`. 
If we don't have browser-like env, we don't render the modal anyways...

What do you think?

keeganstreet:
@diasbruno sounds good, that would require fewer changes. I have just tested that and it will resolve the problem for us. I'll push an update to this branch.

diasbruno (approved):

diasbruno:
Thanks, @keeganstreet!!

diasbruno:
I've updated the title so it's easy to find.

keeganstreet:
Not a problem @diasbruno :-) Thanks for your help!

Do you have a release schedule for npm? Wondering when we'll be able to have this fix for our system or if we need to put in a temporary workaround?

PR #835, opened 2020-10-23
[fixed] add aria-modal attribute

youf-olivier:
As it said in the [comment](https://github.com/reactjs/react-modal/issues/654#issuecomment-672640981) of #654 Safari has no longer problem with aria-modal='true'

Without aria-modal, React Testing library seems have some troubles to find de modal role.

Now, with the current version of the code of React-modal and Safari, the modal seems to be able to be reintroduce.

Acceptance Checklist:
- [x] Documentation (README.md) and examples have been updated as needed. => No Needed
- [x] If this is a code change, a spec testing the functionality has been added. => No Needed
- [x] If the commit message has [changed] or [removed], there is an upgrade path above. => No Needed

### DIFF
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -379,7 +379,7 @@ export default class ModalPortal extends Component {
       onClick: this.handleContentOnClick,
       role: this.props.role,
       "aria-label": this.props.contentLabel,
-      ...this.attributesFromObject("aria", this.props.aria || {}),
+      ...this.attributesFromObject("aria", { modal: true, ...this.props.aria }),
       ...this.attributesFromObject("data", this.props.data || {}),
       "data-testid": this.props.testId
     };


src/components/ModalPortal.js
```diff
@@ -379,6 +379,7 @@ export default class ModalPortal extends Component {
       onClick: this.handleContentOnClick,
       role: this.props.role,
       "aria-label": this.props.contentLabel,
+      "aria-modal": true,
```
diasbruno, line 382: You can set this property using `aria={{ "modal": true }}`. Not sure if it would be ok to set by default.

Any thoughts?

youf-olivier: Yes, I saw this, but it make sense to be here, it's not really a addition, but a basic for a good accessible modal. 

But of course, if the PR is rejected, I have a other solution with the aria additionnal properties, but I have the feeling that is place is by default.

diasbruno: That's what I'm worried about...if this aria attribute, added by default, would conflict with any other accessibility role.

You can find more details here (W3C Modal)[https://www.w3.org/TR/wai-aria-practices/#dialog_modal].

youf-olivier: I based my proposal on the W3C recommandation, on the same link you can find a few lines below : 

https://www.w3.org/TR/wai-aria-practices/#dialog_roles_states_props

```
The dialog container element has aria-modal set to true.
```

But as you wish :)

youf-olivier: If you reject the PR please close it, that I'll have to make another PR in my Design System to correct it by my own

diasbruno: You don't need to be a pessimist. :stuck_out_tongue_closed_eyes: 

Maybe if you move this to the line below the patch, people can control this attribute (not sure why someone would do that, but better to keep it editable).

What do you think?

src/components/ModalPortal.js
```diff
@@ -379,6 +379,7 @@ export default class ModalPortal extends Component {
...
       ...this.attributesFromObject("aria", this.props.aria || {}),
```
diasbruno, line 383:
```suggestion
      ...this.attributesFromObject("aria",  { modal: true, ...this.props.aria}),
```

This would allow the user to specify in the aria attribute.

youf-olivier: Ah ok. Yes it's a good idea :)

diasbruno (approved):
🎉

diasbruno:
@youf-olivier Thanks for having a look at this.

PR #861, opened 2021-03-31
[added] support Array, HTMLCollection and NodeList values for appElement

eemeli:
**Changes proposed:**

- Hide multiple app elements, not just one
- Accept Array, HTMLCollection & NodeList values for the `appElement` prop

While working with an application that uses more than one top-level element, it became obvious that providing a fully accessible experience via react-modal is challenging, as it's currently only able to hide one app element.

The fix for this is pretty simple, and should have no effect on existing users. The accepted shape of the `appElement` value now also includes Array, HTMLCollection (returned by `document.getElementsBy*`) and NodeList (returned by `document.querySelectorAll`).

As a result of this change, it becomes easier to use react-modal to provide a fully accessible experience.

**Acceptance Checklist:**

### DIFF
--- a/docs/accessibility/index.md
+++ b/docs/accessibility/index.md
@@ -23,6 +23,9 @@ rewritten:
 Modal.setAppElement(document.getElementById('root'));
 ```
 
+Using a selector that matches multiple elements or passing a list of DOM
+elements will hide all of the elements.
+
 If you are already applying the `aria-hidden` attribute to your app content
 through other means, you can pass the `ariaHideApp={false}` prop to your modal
 to avoid getting a warning that your app element is not specified.
--- a/docs/accessibility/index.md
+++ b/docs/accessibility/index.md
@@ -24,7 +24,10 @@ Modal.setAppElement(document.getElementById('root'));
 ```
 
 Using a selector that matches multiple elements or passing a list of DOM
-elements will hide all of the elements.
+elements will hide all of the elements. Note that this list won't be
+automatically pruned if elements are removed from the DOM, so you may want to
+call `Modal.setAppElement` when any such changes are made, or pass a live
+HTMLCollection as the value.
 
 If you are already applying the `aria-hidden` attribute to your app content
 through other means, you can pass the `ariaHideApp={false}` prop to your modal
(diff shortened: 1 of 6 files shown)


package-lock.json
```diff
@@ -1046,27 +1046,27 @@
       "dev": true
```
diasbruno, line 1046: Theses changes can be discarded...

diasbruno:
Looks great. Bom trabalho!

`Modal.setAppElement()`...I'm worried this can potentially hold dead elements.

Any thoughts?

package-lock.json
eemeli: Bother, I had not intended to include package-lock.json in the PR, will remove.

eemeli:
> `Modal.setAppElement()`...I'm worried this can potentially hold dead elements.

That is a possibility. Proactively getting rid of dead elements would probably require adding something like a `document.body.contains(...)` check within the `validateElement()` call to act as a filter. It would be nice if a WeakSet could be used, but that's not iterable.

Should I add that sort of filter?

diasbruno:
> It would be nice if a WeakSet

Yeah, totally agree. 

I don't know if we should be defensive in this case, or, just write it in the docs warning about it.

docs/accessibility/index.md
```diff
@@ -23,6 +23,9 @@ rewritten:
 Modal.setAppElement(document.getElementById('root'));
 ```
 
+Using a selector that matches multiple elements or passing a list of DOM
+elements will hide all of the elements.
+
```
diasbruno, line 28: Let's add a note about the risk of using `setAppElement`.

eemeli: Done. Also mentioned the possibility of using a HTMLCollection, as that does update its contents whenever the DOM updates.

diasbruno (approved):

diasbruno:
Thanks, @eemeli.

eemeli:
Any chance of getting a release including this sometime soon? 😇

diasbruno:
Sorry for taking soooo long...released version 3.13.1.

PR #862, opened 2021-04-03
Updated README.md

harshit1902:
Just updated one reference URL  in the README.md file.
BTW thanks for such a great repository that saves me every time to write whole HTML, CSS, JS for just a modal.


diasbruno:
Thanks, @harshit1902.

Issue #799, opened 2020-02-11
Ability to disable the focus trap


chaddjohnson:
### Summary:

It would be great if it was possible to disable the focus trap. This would be useful if ever needing to render the modal inline in a page.

I am building an app which lets one build a modal, and it would be useful to turn off the focus trap while in design mode.

JKillian:
@chaddjohnson I think setting `shouldFocusAfterRender={false}` will disable the focus trap. If you want it to still gain focus initially after opening, something like `onAfterOpen={() => this.contentRef.focus()}` might work.

chaddjohnson:
@JKillian Thank you. I wonder if there also is a way to disable the focus trap to enable interaction with the modal without trapping focus.

JKillian:
@chaddjohnson I believe that `shouldFocusAfterRender={false}` _does_ disable the focus trap. But not 100% sure, just what glancing at the code looked like to me!

Vanals:
Having the same issue here! Would be really useful if we could! In order to make it flexible for other dialogs types!

I tried `shouldFocusAfterRender={false}` but doesn't stop the trapping to work. @JKillian

Vanals:
Anyone? :D 
@chaddjohnson  could you explain to me how you have overcome the obstacle?
I can't see how having an overlay on top of the modal should stop the JS logic from trapping the user inside?

Thanks!

chaddjohnson:
An overlay on top prevents focus from ever entering the dialogue via click events. Keyboard tabbing into the modal may still cause focus to be trapped...I haven’t tested that.

bvellacott:
Unfortunately setting `shouldFocusAfterRender={false}` does not disable the focus trap.

bvellacott:
I've created a PR to address this issue: https://github.com/reactjs/react-modal/pull/876

bvellacott:
We have a guide modal in our app. It can be viewed while using the app, but if it receives focus it needs to be closed before the user can transfer focus back to the rest of the app using the tab button. This is a major accessibility concern for us. All the other functionality of react-modal suits the use case very well.

diasbruno:
Please see discussion #881. It'd be nice to have some feedback on this.

PR #876, opened 2021-05-21
Disable focus trap & disable tab trap

bvellacott:
Fixes #[issue number].

Changes proposed:

- add ability to disableFocusTrap
- add ability to disableTabTrap

Acceptance Checklist:

### DIFF
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -273,7 +273,7 @@ export default class ModalPortal extends Component {
   };
 
   handleKeyDown = event => {
-    if (event.keyCode === TAB_KEY) {
+    if (event.keyCode === TAB_KEY && !this.props.disableTabTrap) {
       scopeTab(this.content, event);
     }
 
--- a/src/helpers/bodyTrap.js
+++ b/src/helpers/bodyTrap.js
@@ -53,7 +53,9 @@ function bodyTrap(eventType, openInstances) {
 
   instances = openInstances;
 
-  if (instances.length > 0) {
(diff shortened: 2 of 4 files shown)


bvellacott:
https://github.com/reactjs/react-modal/issues/799

src/helpers/bodyTrap.js
```diff
@@ -53,7 +53,9 @@ function bodyTrap(eventType, openInstances) {
 
   instances = openInstances;
 
-  if (instances.length > 0) {
+  const trapFocus = instances.find(({ props }) => !props.disableFocusTrap);
+
+  if (trapFocus && instances.length > 0) {
```
diasbruno, line 58: What happens if we open one modal with focus trap disabled and one that's enabled?

src/components/ModalPortal.js
```diff
@@ -273,7 +275,7 @@ export default class ModalPortal extends Component {
   };
 
   handleKeyDown = event => {
-    if (event.keyCode === TAB_KEY) {
+    if (event.keyCode === TAB_KEY && !this.props.disableTabTrap) {
```
diasbruno, line 278:
```suggestion
    if (!this.props.disableTabTrap && event.keyCode === TAB_KEY) {
```

src/helpers/bodyTrap.js
bvellacott: Then a focus trap will exist. I thought that is the best way to do it as it gives existing functionality priority over the added functionality. I could add something in the readme to let the user know?

bvellacott: I added a comment in the readme about this

PR #896, opened 2021-08-13
[fixed] Ensure we don't check for hidden on Shadow DOM.

stevematney:
When inside a Shadow DOM, the loop will eventually reach the Shadow DOM element itself, which cannot be checked for visibility. We can continue our upwards check by skipping to the Shadow DOM host.

Changes proposed:

- Skip the Shadow DOM element (`rootNode` when inside a Shadow DOM) when checking for visibility.
- Ensure that we know about every tabbable element, even those within Shadow DOM.

Acceptance Checklist:

### DIFF
--- a/specs/Modal.helpers.spec.js
+++ b/specs/Modal.helpers.spec.js
@@ -1,5 +1,6 @@
 /* eslint-env mocha */
 import "should";
+import "@webcomponents/custom-elements/src/native-shim";
 import tabbable from "../src/helpers/tabbable";
 import "sinon";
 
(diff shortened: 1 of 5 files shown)


stevematney:
@diasbruno I added some specs to this to verify the functionality.

diasbruno:
@stevematney Look really good! Bom trabalho. I'll do my best to review this ASAP.

stevematney:
Cool! Thank you!

specs/Modal.helpers.spec.js
```diff
@@ -15,12 +38,16 @@ export default () => {
...
         elem = document.createElement("div");
         document.body.appendChild(elem);
       });
 
       afterEach(() => {
-        document.body.removeChild(elem);
+        document
+          .querySelectorAll("test-web-component")
+          .forEach(node => node.remove());
+        elem.remove();
```
diasbruno, line 50: Maybe it would be better to split these tests.

diasbruno (changes requested):
See comment.

specs/Modal.helpers.spec.js
stevematney: I separated them into a separate `describe`. Does that work for you?

diasbruno (approved):

stevematney:
@diasbruno we recently discovered that simply using `document.activeElement` does not work in Safari with Web Components inside Modals that have Shadow DOM attached. For this instance, you have to drill down through (potentially) nested `shadowRoot` elements to find the actual element that is active. Without this drilling down, `scopeTab` thinks that elements inside a Shadow DOM are _not_ visible within the modal, even though that's not the case. This breaks tabbing, either redirecting focus to the Modal itself, or to the first non-Web-Component item within the Modal.

This is an active bug for us, so the sooner we can get these changes into a new `react-modal` version, the better. I appreciate all your help on this!

src/helpers/tabbable.js
```diff
@@ -59,5 +65,14 @@ function tabbable(element) {
...
 
 export default function findTabbableDescendants(element) {
-  return [].slice.call(element.querySelectorAll("*"), 0).filter(tabbable);
+  const descendants = [].slice
+    .call(element.querySelectorAll("*"), 0)
+    .reduce(
+      (finished, el) => [
+        ...finished,
+        ...(!el.shadowRoot ? [el] : findTabbableDescendants(el.shadowRoot))
+      ],
```
diasbruno, line 74:
```suggestion
      (finished, el) => finished.concat(
        !el.shadowRoot ? [el] : findTabbableDescendants(el.shadowRoot)
      ),
```

diasbruno:
Will `getActiveElement` do the job for this case?

stevematney:
> Will `getActiveElement` do the job in this case?

I think you're asking if the new `getActiveElement` function in these changes will solve the problem of finding the `activeElement` within Shadow DOM. If that is what you're asking, then yes. It will recurse through nested `activeElement` until it finds the one that does not have a `shadowRoot`.

https://github.com/reactjs/react-modal/blob/3aa914fbb14c4011911d4c2167ab1015d0598f03/src/helpers/scopeTab.js#L3-L7

diasbruno:
Nice! Released version `3.14.4`.

stevematney:
> Nice! Released version `3.14.4`. 🎉

@diasbruno thanks so much for jumping on that quickly!

diasbruno:
Nah..no problem, bro! 
Let me know if something happens in this release so we can rollback.

PR #906, opened 2021-10-11
[fixed] element with display 'contents' is visible and is tabbable.

### LINKED ISSUE #905: Tabbing not possible inside display: contents
### Summary:
Within a modal, it is not possible to tab through elements inside an element with display: contents

### Steps to reproduce:

 1. Create a modal with an element that has display: contents
 2. Put focus-able elements inside that element (eg inputs)
 3. Try to tab through the inputs
 4. No focus change on tab

### Expected behavior:
Focus switches to the next input element on tab

### Link to example of issue:
See https://codesandbox.io/s/flamboyant-sunset-t3pbm
Open the dialog, then try to tab through the inputs - it won't work. Then change display: "contents" to display: "block" - tabbing no works.

### Additional notes:
This is caused by the fact that elements with display: "contents" have a size of 0x0 and tabbable assumes the content is invisible:
https://github.com/reactjs/react-modal/blob/fb6bab5e7fda5b616efca201cf9b68df21e592f9/src/helpers/tabbable.js#L15-L34

It can be fixed by checking for display: contents, like so:
```
diff --git a/node_modules/react-modal/lib/helpers/tabbable.js b/node_modules/react-modal/lib/helpers/tabbable.js
index c72bf54..0e8a07a 100644
--- a/node_modules/react-modal/lib/helpers/tabbable.js
+++ b/node_modules/react-modal/lib/helpers/tabbable.js
@@ -27,6 +27,10 @@ function hidesContents(element) {
   try {
     // Otherwise we need to check some styles
     var style = window.getComputedStyle(element);
+    if(style.getPropertyValue("display") === "contents"){
+      // display === "contents" does not hide its contents even if its size is (always) 0x0
+      return false;
+    }
     return zeroSize ? style.getPropertyValue("overflow") !== "visible" ||
     // if 'overflow: visible' set, check if there is actually any overflow
     element.scrollWidth <= 0 && element.scrollHeight <= 0 : style.getPropertyValue("display") == "none";
```

diasbruno:
Fixes #905.

Changes proposed:

- element with display 'contents' is visible and is tabbable.

- None

Acceptance Checklist:

### DIFF
--- a/src/helpers/tabbable.js
+++ b/src/helpers/tabbable.js
@@ -10,7 +10,16 @@
  * http://api.jqueryui.com/category/ui-core/
  */
 
+const DISPLAY_NONE = "none";
+cosnt DISPLAY_CONTENTS = "contents";
+
 const tabbableNode = /input|select|textarea|button|object/;
+
+function hasElementOverflow(element, style) {
+  return style.getPropertyValue("overflow") !== "visible" ||
+    // if 'overflow: visible' set, check if there is actually any overflow
+    (element.scrollWidth <= 0 && element.scrollHeight <= 0);
+}
 
 function hidesContents(element) {
   const zeroSize = element.offsetWidth <= 0 && element.offsetHeight <= 0;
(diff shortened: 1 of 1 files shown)


diasbruno:
@galexandrade That's great! I believe the only thing missing here is to add a tests. You can take this PR and improve it.

galexandrade:
@diasbruno @aloker I checked out the branch and tried to add some tests to cover the scenario but since the problem only happens on Safari, it is hard to simulate the same scenario on ReactDOM. 

This is the problem I would like to have fixed here:
If there is an element that has the `display: contents` style applied, on Safari, it won't tab through the elements. The reason I am using `display: contents` is that I want my `form` tag to wrap the modal content and footer so the modal buttons are part of the form. However, it is messing with the modal layout so, I use the `display: contents` so the browser will "ignore" that form tag when painting the screen.

https://user-images.githubusercontent.com/24475665/195448097-b2de8a8c-e7a5-41f1-a1d5-c999966020fa.mov

That works fine, except for Safari where it won't tab through the elements. With this PR, the issue is fixed.

I'd like to offer help to move this PR forward. Please, how can I help to get this PR merged? 🙏

diasbruno:
@galexandrade There are a few comments on the PR. You can take this patch and improve on it.

diasbruno:
#969

PR #908, opened 2021-10-20
[added] tabbable support for iframes

jaksenko:
Iframes are, just like a button, anchor or inputs, an interactive element that should be focusable in modal. For example I play youtube videos in modals. Keyboard users should be able to pause video (or trigger any other action) as well.

Changes proposed:

- add `iframe` to list of tabbable elements

N/A

Acceptance Checklist:

### DIFF
--- a/src/helpers/tabbable.js
+++ b/src/helpers/tabbable.js
@@ -10,7 +10,7 @@
  * http://api.jqueryui.com/category/ui-core/
  */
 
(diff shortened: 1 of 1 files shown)


diasbruno:
Obrigado, @jaksenko!

jaksenko:
Thank you for merging! Do you think you will be able to publish it to npm so we can use it in our projects? 🙂

PR #940, opened 2022-04-20
[changed] add stopPropagation at handleOverlayOnClick

### LINKED ISSUE #939: handleOverlayOnClick is not stopping click event propagation
### Summary:

Modal is propagating the click event to the parent component

### Steps to reproduce:

See the issue example link

### Expected behavior:

Modal must not propagate the click event to the parent component.

### Link to example of issue:

https://stackblitz.com/edit/react-wa4nwp?file=src/App.js

joaoGabriel55:
Fixes #939.

Changes proposed:

- Add stopPropagation at handleOverlayOnClick

- None

Acceptance Checklist:

### DIFF
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -285,6 +285,8 @@ export default class ModalPortal extends Component {
   };
 
   handleOverlayOnClick = event => {
+    event.stopPropagation();
+
     if (this.shouldClose === null) {
       this.shouldClose = true;
     }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,16907 @@
 {
   "name": "react-modal",
-  "version": "3.14.4",
-  "lockfileVersion": 1,
+  "version": "3.15.1",
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "react-modal",
+      "version": "3.15.1",
+      "license": "MIT",
+      "dependencies": {
+        "exenv": "^1.2.0",
+        "prop-types": "^15.7.2",
+        "react-lifecycles-compat": "^3.0.0",
+        "warning": "^4.0.3"
+      },
+      "devDependencies": {
+        "@webcomponents/custom-elements": "^1.5.0",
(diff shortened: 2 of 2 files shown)


src/components/ModalPortal.js
```diff
@@ -285,6 +285,8 @@ export default class ModalPortal extends Component {
   };
 
   handleOverlayOnClick = event => {
+    event.stopPropagation();
```
smcenlly, line 288: I came across this pull request as my events aren't propagating right now when I want them to (I'm doing something wrong in my app). My scenario is that I have a transparent overlay and `shouldCloseOnOverlayClick` is set. I want the modal to close and underlying control to be clicked.

I don't think that `event.stopPropagation()` should always happen; it should be opt-in (or a setting to control the behavior). With this change, I believe it's possible to break existing use of `react-modal` that relies on this behavior.

joaoGabriel55: Thanks for the review @smcenlly. I will take a look

diasbruno: We can add the stopPropagation there...we need to be careful here to only stop if the event is target is the overlay.

joaoGabriel55: I fixed this, in this commit: https://github.com/reactjs/react-modal/pull/940/commits/4dd8ee3cf9e09606bd3170e97712447756e39480

diasbruno: Muito bem...um potiguar por aqui!!

I think we are close to get this ready...can this `event` not be related to the overlay? I'm not sure if that is a valid case, but it would be nice to check it.

TheRakeshPurohit:
This needs to be merged and released if this allows handling overlay click events.

src/components/ModalPortal.js
TheRakeshPurohit: This needs to be merged and released if this allows handling overlay click events.

underscore05:
I came across the same issue. What I've done is I wrap the Modal component in `<div onClick={(e) => e.stopPropagation()}>`. This way it will stop the event on that wrapper.

PR #961, opened 2022-08-21
Added close modal on right click

ZainGulbaz:
Fixes #837 .

Changes proposed:

- Added feature to close modal on right click on overlay
- Use props shouldCloseOnOverlayRightClick={true} to enable this feature
- By default it is false
-Its functionality is same as prop shouldCloseOnOverlayClick
` <Modal  shouldCloseOnOverlayRightClick={true}>`

Acceptance Checklist:
- [ ] The commit message follows the guidelines in `CONTRIBUTING.md`.
- [ ] Documentation (README.md) and examples have been updated as needed.
- [ ] If this is a code change, a spec testing the functionality has been added.
- [ ] If the commit message has [changed] or [removed], there is an upgrade path above.

PR #963, opened 2022-08-27
[Added] Modal Close on right click at overlay

ZainGulbaz:
Fixes #[837].

Changes proposed:

-Added a prop of function type 'onOverlayRightClick' that calls a function when ever we right click on modal's overlay.
-The user can then change modal's state isOpen to false, if he wants to clocse the modal on right click at overlay
-`   <Modal
          className="Modal__Bootstrap modal-dialog"
          isOpen={this.state.modalIsOpen}
          onRequestClose={this.handleModalCloseRequest}
          **onOverlayRightClick={this.handleModalCloseRequest}**
        >`

Acceptance Checklist:

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -52,6 +52,12 @@ import ReactModal from 'react-modal';
   /* Function that will be run when the modal is requested
      to be closed (either by clicking on overlay or pressing ESC).
      Note: It is not called if isOpen is changed by other means. */}
+     
+  onOverlayRightClick={
+    handleRequestCloseFunc
+    /* Function that will be run when the modal is requested
+     to be closed (either by right clicking on overlay or pressing ESC).
+     Note: It is not called if isOpen is changed by other means. */}
 
   closeTimeoutMS={
     0
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -66,6 +66,7 @@ class Modal extends Component {
     ]),
     onAfterOpen: PropTypes.func,
     onRequestClose: PropTypes.func,
+    onOverlayRightClick: PropTypes.func,
     closeTimeoutMS: PropTypes.number,
     ariaHideApp: PropTypes.bool,
     shouldFocusAfterRender: PropTypes.bool,
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -55,6 +55,7 @@ export default class ModalPortal extends Component {
     onAfterOpen: PropTypes.func,
     onAfterClose: PropTypes.func,
     onRequestClose: PropTypes.func,
+    onOverlayRightClick: PropTypes.func,
     closeTimeoutMS: PropTypes.number,
     shouldFocusAfterRender: PropTypes.bool,
     shouldCloseOnOverlayClick: PropTypes.bool,
@@ -299,6 +300,15 @@ export default class ModalPortal extends Component {
     this.shouldClose = null;
   };
 
+  handleOverlayRightClick = (event) => {
+    if (this.shouldClose === null) this.shouldClose = true;
+    else if (!this.shouldClose) this.shouldClose = null;
+    if (this.shouldClose) {
+      event.preventDefault();
+      this.props.onOverlayRightClick(event);
+    }
+  };
+
   handleContentOnMouseUp = () => {
     this.shouldClose = false;
   };
@@ -375,7 +385,8 @@ export default class ModalPortal extends Component {
       className: this.buildClassName("overlay", overlayClassName),
       style: { ...overlayStyles, ...this.props.style.overlay },
       onClick: this.handleOverlayOnClick,
-      onMouseDown: this.handleOverlayOnMouseDown
+      onMouseDown: this.handleOverlayOnMouseDown,
+      onContextMenu: this.handleOverlayRightClick,
     };
 
     const contentProps = {
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ function App() {
         isOpen={modalIsOpen}
         onAfterOpen={afterOpenModal}
         onRequestClose={closeModal}
+        onOverlayRightClick={closeModal}
         style={customStyles}
         contentLabel="Example Modal"
       >
--- a/examples/bootstrap/app.js
+++ b/examples/bootstrap/app.js
@@ -39,6 +39,7 @@ class App extends Component {
           closeTimeoutMS={150}
           isOpen={this.state.modalIsOpen}
           onRequestClose={this.handleModalCloseRequest}
+          onOverlayRightClick={this.handleModalCloseRequest}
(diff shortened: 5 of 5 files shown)


src/components/ModalPortal.js
```diff
@@ -299,6 +300,15 @@ export default class ModalPortal extends Component {
...
   };
 
+  handleOverlayRightClick = (event) => {
+    if (this.shouldClose === null) this.shouldClose = true;
+    else if (!this.shouldClose) this.shouldClose = null;
+    if (this.shouldClose) {
+      event.preventDefault();
+      this.props.onOverlayRightClick(event);
+    }
+  };
```
diasbruno, line 310: hmmm...:thinking: I don't think we can initiate the closing process here. It must happend on the user side. So, the only thing we can do here is to forward the callback the user defined for `onOverlayRightClick`.

diasbruno: If someone want to just use the right click for other purpose, this will conflict.

diasbruno:
We are going in the right direction, just made some comments. It must be the last change. 
Obrigado pelo PR.

diasbruno:
Awesome!! :tada: 

Can you include a test for this, please?

src/components/ModalPortal.js
ZainGulbaz: @diasbruno Kindly check my latest commit. I think now it's okay.

ZainGulbaz:
@diasbruno I have add latest commit of example test case. Kindly let me know, if you have any suggestions.

diasbruno:
It looks great, but we also need to add a unit testing for this feature.  Let me know if you need anything...

ZainGulbaz:
Is there any unit testing already done?

diasbruno:
All unit testing are here https://github.com/reactjs/react-modal/tree/master/specs

docs/index.md
```diff
@@ -52,6 +52,12 @@ import ReactModal from 'react-modal';
   /* Function that will be run when the modal is requested
      to be closed (either by clicking on overlay or pressing ESC).
      Note: It is not called if isOpen is changed by other means. */}
+     
+  onOverlayRightClick={
```
doeg, line 56: I would propose `onOverlayContextMenu` as a better name for this prop, per @diasbruno's suggestion [here](https://github.com/reactjs/react-modal/pull/961#issuecomment-1222428392).

We are passing this to React's `onContextMenu` event handler, which maps to the `contextmenu` event (see [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event)):

> The contextmenu event fires when the user attempts to open a context menu. This event is typically triggered by clicking the right mouse button, or by pressing the context menu key.

Some alternate ways to trigger the `contextmenu` event beyond right click:
- `ctrl` + click on MacOS 
- `shift+f10` on Windows
- `fn+12` on MacOS (with "Alternate Pointer Events" enabled)
 
...And some screen readers have their own triggers, too.

docs/index.md
```diff
@@ -52,6 +52,12 @@ import ReactModal from 'react-modal';
...
+    handleRequestCloseFunc
+    /* Function that will be run when the modal is requested
+     to be closed (either by right clicking on overlay or pressing ESC).
+     Note: It is not called if isOpen is changed by other means. */}
```
doeg, line 60: This comment is out of date given that 06c1c0451230f08966b63fc174160ff8a20cf4f2 changed the prop to be a pass-through to `onContextMenu` instead of actually closing the modal. It is also not triggered by the `esc` key.

I would suggest a straightforward comment like:

> Function that will be run when the context menu is triggered on the overlay (e.g., by right clicking or by pressing the context menu).

src/components/ModalPortal.js
```diff
@@ -299,6 +300,14 @@ export default class ModalPortal extends Component {
     this.shouldClose = null;
   };
 
+  handleOverlayRightClick = event => {    
+      if(event.target == this.overlay)
+      {
+        event.preventDefault();
```
doeg, line 306: We may want to allow the `this.props.onOverlayRightClick` function to call `event.preventDefault()` if they want to, for more flexibility. @diasbruno what do you think? 🤔

examples/basic/modal_close_right_click/index.js
```diff
@@ -0,0 +1,50 @@
...
+      isOpen: false,
+    };
+  }
+
+  toggleModal = (event) => {
+    this.setState({ isOpen: !this.state.isOpen });
+  };
+  handleRightClick = (event) => {
+    this.setState({ isOpen: false });
+  };
```
doeg, line 17: Nit: missing a newline in between these two functions.

examples/basic/modal_close_right_click/index.js
```diff
@@ -0,0 +1,50 @@
...
+          className="btn btn-primary"
+          onClick={this.toggleModal}
+        >
+          Open Modal
+        </button>
+        <Modal
+          isOpen={this.state.isOpen}
+          onOverlayRightClick={this.handleRightClick}
+        >
+          <h1>Click Right on the Overlay to close the modal</h1>
```
doeg, line 33: Nit: rephrase to "Right click on the overlay to close the modal"

src/components/Modal.js
```diff
@@ -100,7 +101,8 @@ class Modal extends Component {
     preventScroll: false,
     parentSelector: () => document.body,
     overlayElement: (props, contentEl) => <div {...props}>{contentEl}</div>,
-    contentElement: (props, children) => <div {...props}>{children}</div>
+    contentElement: (props, children) => <div {...props}>{children}</div>,
+    onOverlayRightClick: ()=>null
```
doeg, line 105: Nit: missing whitespace:

```jsx
onOverlayRightClick: () => null
```

doeg (commented):
I realize this PR is quite old now, but it seems like it could be a useful feature! I left some review comments. :)

docs/index.md
doeg: Note that if we do decide to rename the prop, it may be worth updating the commits/PR title so that the wording in the changelog is accurate.

diasbruno: > fn+12 on MacOS (with "Alternate Pointer Events" enabled)

This one I didn't know. :)

diasbruno: Yep, agree. Better to keep the names consistent.

src/components/ModalPortal.js
diasbruno: Better for react-modal to not interfere in the event propagation. It would be better to let the user to handles it. So, better to just forward the event if the handle is present.

diasbruno:
@honeymaro It depends. 

You should use `overlayRef` only if you need to do something with the reference. 
In this case, having 2 options is not a terrible idea.

Issue #964, opened 2022-09-12
bodyOpenClassName is added to incorrect document


leoc4e:
### Steps to reproduce:

 1.  use reactModal inside iframe by using `parentSelector`
 2. add className to body by using `bodyOpenClassName`
 3. className is added to rootDocument body

### Expected behavior:
className should add to the ownerDocument of the modal instead

PR #965, opened 2022-09-12
[fixed]: css class added to root document instead of modal ownerDocument

leoc4e:
Fixes [#[964]](https://github.com/reactjs/react-modal/issues/964)
I am trying to open the modal inside an iframe, but the `bodyOpenClassName` would be applied to the rootDocument instead.
I would like to use the parentSelector to determine where the `bodyOpenClassName` should be added instead of just using `document.body`

Changes proposed:

- `htmlOpenClassName` add to the modal ownerDocument html instead
- `bodyOpenClassName` add to the modal ownerDocument body instead
- Update doc 

Acceptance Checklist:

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -90,7 +90,7 @@ import ReactModal from 'react-modal';
 
   bodyOpenClassName={
     "ReactModal__Body--open"
-  /* String className to be applied to the document.body
+  /* String className to be applied to the modal ownerDocument.body
      (must be a constant string).
      This attribute when set as `null` doesn't add any class
      to document.body.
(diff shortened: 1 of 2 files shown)


diasbruno:
LEGIT.

leoc4e:
@diasbruno 
Is there anything I should do for this PR to move forward?

diasbruno:
I didn't have time to look at it, but one thing is I don't know how old is `ownerDocument`. If you can have look on it to see if it is safe to use, then ok...

leoc4e:
I think it should be safe to use,
image
https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument
https://caniuse.com/?search=ownerDocument
it is supported even for IE 6 in 2002
for android it is not supported until 4.4 (2014)

And since I am using `(parentSelector && parentSelector().ownerDocument) || document`
it should fallback to use the `document` if `ownerDocument` is `undefined`/`null`🤞

diasbruno:
Great. Thanks for the the PR e bom trabalho!

PR #966, opened 2022-09-21
Add overlay/content example

arist0tl3:
Added an example I used to animate modal content in and apply an opacity transition to the overlay. If this is interesting I can look into adding to the `/examples` as well.

Acceptance Checklist:
- [n/a] Documentation (README.md) and examples have been updated as needed.
- [n/a] If this is a code change, a spec testing the functionality has been added.
- [n/a] If the commit message has [changed] or [removed], there is an upgrade path above.

### DIFF
--- a/docs/styles/transitions.md
+++ b/docs/styles/transitions.md
@@ -1,7 +1,10 @@
 Using [CSS classes](classes.md), it is possible to implement transitions for
-when the modal is opened or closed.  By placing the following CSS somewhere in
-your project's styles, you can make the modal content fade in when it is opened
-and fade out when it is closed:
+when the modal is opened or closed.  
+
+#### Basic overlay styling
+
+By placing the following CSS somewhere in your project's styles, you can make the 
+modal content fade in when it is opened and fade out when it is closed:
 
 ```css
 .ReactModal__Overlay {
@@ -18,12 +21,13 @@ and fade out when it is closed:
 }
 ```
 
-
 The above example will apply the fade transition globally, affecting all modals
 whose `afterOpen` and `beforeClose` classes have not been set via the
 `className` prop.  To apply the transition to one modal only, you can change
 the above class names and pass an object to your modal's `className` prop as
 described in the [previous section](classes.md).
+
+#### Gracefully closing the modal
 
 In order for the fade transition to work, you need to inform the `<Modal />` about the transition time required for the animation.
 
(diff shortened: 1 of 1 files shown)


diasbruno:
LEGIT. Obrigado, @arist0tl3!

diasbruno:
The style is the "default style" that we use on the examples, right?

arist0tl3:
@diasbruno 

I think I understand what you were getting at -- just updated the example to use the default styles and note that `transform` will need to be removed in order for this example to work properly.

diasbruno:
Great. Is there a bug on the styles of the examples?

arist0tl3:
No, but if I try to do this:

```
.ReactModal__Content {
  transform: translate(-50%, 150%);
  transition: transform 800ms ease-in-out;
}

.ReactModal__Content--after-open {
  transform: translate(-50%, -50%);
}

.ReactModal__Content--before-close {
  transform: translate(-50%, 150%);
}
```

the inline style will take precedence. It felt better to suggest removing that inline style and define it instead as a css rule vs keeping it as is and adding `!important` to the rules above.

docs/styles/transitions.md
```diff
@@ -72,6 +76,68 @@ Instead of this
...
+  <Modal
+    style={{
+      overlay: { background: 'rgba(0, 0, 0, 0.3)' },
+      content: {
+        top: '50%',
+        left: '50%',
+        right: 'auto',
+        bottom: 'auto',
+        marginRight: '-50%',
+        // transform: 'translate(-50%, -50%)', This will override the css values if not removed
```
diasbruno, line 96: Maybe we can move this note after the first paragraph to warn about the precedence.

docs/styles/transitions.md
```diff
@@ -72,6 +76,68 @@ Instead of this
 }
 ```
 
+#### Styling the overlay and content
+
+It is possible to independently style the transitions for the modal 
+overlay and modal content. The example below will fade the overlay 
+in and slide the modal content on-screen from the bottom of the page:
```
diasbruno, line 83:
```suggestion
It is possible to independently style the transitions for the modal 
overlay and modal content. 

{about precendence}

The example below will fade the overlay 
in and slide the modal content on-screen from the bottom of the page:
```

arist0tl3:
Took a stab at wording it; lmk if it does/doesn't make sense lol

docs/styles/transitions.md
```diff
@@ -72,6 +76,75 @@ Instead of this
...
 
+#### Styling the overlay and content
+
+It is possible to independently style the transitions for the modal 
+overlay and modal content. 
+
+Keep in mind that inline styles will take precedence over those defined 
+as CSS rules. If you want to apply a transition to one of the default styles,
+such as `transform`, you can remove it from the inline definition and define
+it instead as a CSS rule.
```
diasbruno, line 87:
```suggestion
Keep in mind that inline styles will take precedence over those defined 
as CSS rules. In case you need to use both at the same time, you'll need to use the css `!important`.
```

docs/styles/transitions.md
```diff
@@ -72,6 +76,75 @@ Instead of this
...
+  <Modal
+    style={{
+      overlay: { background: 'rgba(0, 0, 0, 0.3)' },
+      content: {
+        top: '50%',
+        left: '50%',
+        right: 'auto',
+        bottom: 'auto',
+        marginRight: '-50%',
+        // transform: 'translate(-50%, -50%)',
```
diasbruno, line 103:
```suggestion
        // transform: 'translate(-50%, -50%)', if you use inline styles instead of css 
```

docs/styles/transitions.md
diasbruno: Just made more general because it will happen to other properties as well.

docs/styles/transitions.md
diasbruno: This line is to clarify why this line is commented.

diasbruno:
Everything looks good. Thanks so much for taking the time to help with the docs. 
Grande Lavoro. :tada: 
Grazie mille.

PR #1012, opened 2023-03-31
[chore] update readme and docs demos

### LINKED ISSUE #955: Accessibility issues with the codepen demos
### Summary:

Many of the demos currently listed have accessibility issues:

* [Minimal example](https://codepen.io/claydiffrient/pen/KNxgav) - once the modal is opened, `aria-hidden="true"` is set on the `<body>`. this also hides the modal itself (which is a child element of the `<body>`) from screen readers; when the modal is closed, focus is not returned explicitly to the "Trigger modal" button
* [Using setAppElement](https://codepen.io/claydiffrient/pen/ENegGJ) - same issues as the minimal example
* [Using inline styles](https://codepen.io/claydiffrient/pen/ZBmyKz) - when the modal is closed, focus is not returned explicitly to the "Trigger modal" button
* [Customizing the default styles](https://codepen.io/claydiffrient/pen/pNXgqQ) - while the `#1 modal` is fine, when triggering `#2 modal` the `<body>` is once again hidden with `aria-hidden="true"` and, as a result, the `#2 modal` is also hidden from screen readers

### Steps to reproduce:

 1. Use a screen reader
 2. Open the codepen demos
 3. Open/close the various modals

### Expected behavior:

* don't set `aria-hidden="true"` on the `<body>` itself, as this will hide *everything* from screen readers, including the dialog
* make sure that focus is returned to the button that first opened the modal

Review how the working demos like [Using inline styles](https://codepen.io/claydiffrient/pen/ZBmyKz) do it, and apply the same behavior to the broken demos.

neilhsmith:
Closes #955.

Changes proposed:

- Update the readme and doc's demos. Updating to a modern version of react-modal fixes the accessibility issues noted in the issue automatically.
- Remove the `setAppElement` example. A warning is now logged when the app element is not set so removed this one and instead setAppElement in all other demos.
- Add a new styled-components example.

The demos themselves have remained largely unchanged apart from converting to functional components. They are still styled simply and do not set any extra props. All demos are stored [in this CodePen collection](https://codepen.io/collection/jbxNWO).

Acceptance Checklist:

### DIFF
--- a/docs/examples/set_app_element.md
+++ b/docs/examples/set_app_element.md
@@ -1,7 +1,0 @@
-# Using setAppElement
-
-This example shows how to use setAppElement to properly hide your application from screenreaders and other assistive technologies while the modal is open.
-
-You'll notice in this example that the aria-hidden attribute is applied to the #main div rather than the document body.
-
-[setAppElement example](https://codepen.io/claydiffrient/pen/ENegGJ)
--- a/docs/examples/styled_components.md
+++ b/docs/examples/styled_components.md
@@ -1,0 +1,5 @@
+# styled-components
+
+To use styled-components you must override the default inline styles by providing a className and/or overlayClassName. You can then provide your own styled component to the contentElement and overlayElement props.
+
+[styled-components example](https://codepen.io/neilhsmith/pen/gOdJGxp?editors=1011)
--- a/README.md
+++ b/README.md
@@ -122,10 +122,10 @@ local development server using `npm start` or `yarn run start`.
 There are several demos hosted on [CodePen](https://codepen.io) which
 demonstrate various features of react-modal:
 
-* [Minimal example](https://codepen.io/claydiffrient/pen/KNxgav)
-* [Using setAppElement](https://codepen.io/claydiffrient/pen/ENegGJ)
-* [Using onRequestClose](https://codepen.io/claydiffrient/pen/KNjVBx)
-* [Using shouldCloseOnOverlayClick](https://codepen.io/claydiffrient/pen/woLzwo)
-* [Using inline styles](https://codepen.io/claydiffrient/pen/ZBmyKz)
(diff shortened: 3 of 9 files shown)


docs/examples/styled_components.md
```diff
@@ -0,0 +1,5 @@
+# styled-components
```
diasbruno, line 1: @neilhsmith +1 I was expecting an exemple with styled-components for a long time. Thanks!

docs/examples/set_app_element.md
```diff
@@ -1,7 +0,0 @@
-# Using setAppElement
```
diasbruno, line 1: Why this was removed?

diasbruno:
> Remove the setAppElement example. A warning is now logged when the app element is not set so removed this one and instead setAppElement in all other demos.

This needs to come back. `setAppElement` has been a hard feature to explain to users.

docs/examples/set_app_element.md
neilhsmith: Maybe I'm missing historical context but I removed it because it feels more of a requirement than a suggestion since a large warning is logged to the console when it's omitted and since every other example already included it. But if it's needed, then sure, I'll add it back in just a few.

diasbruno: Maybe the explaination is bad. `setAppElement` is a feature that is misused all the time.
Problem is that users always place the setAppElement together with the modal utilization, 
which causes problems when testing and others (i'll try to reference them).
Suggestions are welcome.

neilhsmith: OK, added back in. No strong suggestions, I just assumed `setAppElement` wasn't as strongly recommended in the past or something and might not be needed today.

diasbruno: Actually `setAppElement` must be used. Historically the default element that react-modal uses to add the `aria-hidden` attribute is wrong. By default it will use the `document.body` which will turn the whole page inaccessible.

PR #1018, opened 2023-04-24
[fixed]:1015 - Fixed closing by click on layout

### LINKED ISSUE #1015: Modal doesn't close on the first click on the overlay after click on elements containing event.stopPropagation() inside modal
### Summary:
When opening a second modal from within a first modal, closing the second modal results in the first modal not closing on the first click on the overlay. Instead, it requires a second click to close. This issue is similar to issue #958. However, the solution provided there is not really suitable for our cross-platform project, which uses components from react-native, and I would like to avoid creating separate button components for the web modal, in which event.stopPropagation() would be removed, in order to achieve the correct behavior.

### Steps to reproduce:
1. Open the provided code sandbox: https://codesandbox.io/s/react-modal-layout-issue-xgfsvk
2. Click on the "Open Modal" button to open the first modal.
3. Click on the icon in the top left corner of the first modal to open the second modal.
4. Close the second modal by clicking on the overlay or close button.
5. Attempt to close the first modal by clicking on the overlay.

### Expected behavior:
The first modal should close after the first click on the overlay.

### Link to example of issue:
https://codesandbox.io/s/react-modal-layout-issue-xgfsvk

### Additional notes:
This issue also occurs when closing a modal by clicking on the close button, then reopening it and attempting to close it by clicking on the overlay. The modal only closes after the second click, not the first. This is a kind of standard behavior, so you can reproduce the problem for both the first and second modals. Below are the steps to reproduce the problem for the first modal.

### Steps to reproduce:
1. Open the provided code sandbox: https://codesandbox.io/s/react-modal-layout-issue-xgfsvk
2. Click on the "Open Modal" button to open the first modal.
3. Click on the icon in the top right corner of the first modal to close it.
4. Click on the "Open Modal" button again to open the first modal.
5. Attempt to close the first modal by clicking on the overlay.

ErikDeviashin:
Fixes #1015

Changes proposed:

- Fix incorrect behavior when you need to click twice on the overlay to close modal window if you click on something with event.stopPropagation() inside content block and add a test case for that.
- Fix test cases for the overlay MD -> content MU && content MD -> overlay MU cases because they don't work correctly without the "click" itself.

Upgrade Path (for changed or removed APIs): -

Acceptance Checklist:

### DIFF
--- a/specs/Modal.events.spec.js
+++ b/specs/Modal.events.spec.js
@@ -222,7 +222,7 @@ export default () => {
       withModal(props, null, modal => {
         mouseDownAt(moverlay(modal));
         mouseUpAt(mcontent(modal));
-        clickAt(mcontent(modal));
+        clickAt(moverlay(modal));
         requestCloseCallback.called.should.not.be.ok();
       });
     });
--- a/specs/Modal.events.spec.js
+++ b/specs/Modal.events.spec.js
@@ -222,6 +222,7 @@ export default () => {
       withModal(props, null, modal => {
         mouseDownAt(moverlay(modal));
         mouseUpAt(mcontent(modal));
+        clickAt(mcontent(modal));
         requestCloseCallback.called.should.not.be.ok();
       });
     });
@@ -236,8 +237,39 @@ export default () => {
       withModal(props, null, modal => {
         mouseDownAt(mcontent(modal));
         mouseUpAt(moverlay(modal));
+        clickAt(moverlay(modal));
         requestCloseCallback.called.should.not.be.ok();
       });
+    });
+
+    it("click on button containing stopPropagation inside modal shouldn't block closing", () => {
+      const requestCloseCallback = sinon.spy();
+      let innerButton = null;
+      let innerButtonRef = ref => {
+        innerButton = ref;
+      };
+      function click(event) {
+        event.stopPropagation();
+      }
+
+      withModal(
+        {
+          isOpen: true,
+          onRequestClose: requestCloseCallback
+        },
+        <button ref={innerButtonRef} onClick={click} />,
+        modal => {
+          // imitate regular click with all mouse events on button inside modal
+          mouseDownAt(innerButton);
+          mouseUpAt(innerButton);
+          clickAt(innerButton);
+          // imitate regular click with all mouse events on modal overlay
+          mouseDownAt(moverlay(modal));
+          mouseUpAt(moverlay(modal));
+          clickAt(moverlay(modal));
+          requestCloseCallback.called.should.be.ok();
+        }
+      );
     });
   });
 
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -322,13 +322,15 @@ export default class ModalPortal extends Component {
     if (!this.props.shouldCloseOnOverlayClick && event.target == this.overlay) {
       event.preventDefault();
     }
+    this.shouldClose = null;
   };
 
   handleContentOnClick = () => {
     this.shouldClose = false;
   };
 
-  handleContentOnMouseDown = () => {
+  handleContentOnMouseDown = event => {
+    event.stopPropagation();
     this.shouldClose = false;
   };
 


specs/Modal.events.spec.js
```diff
@@ -222,6 +222,7 @@ export default () => {
       withModal(props, null, modal => {
         mouseDownAt(moverlay(modal));
         mouseUpAt(mcontent(modal));
+        clickAt(mcontent(modal));
```
diasbruno, line 225: Here, mouse down and click at has the effect of a double click.

specs/Modal.events.spec.js
```diff
@@ -236,9 +237,40 @@ export default () => {
       withModal(props, null, modal => {
         mouseDownAt(mcontent(modal));
         mouseUpAt(moverlay(modal));
+        clickAt(moverlay(modal));
```
diasbruno, line 240: Same here.

diasbruno:
@ErikDeviashin Can you please rebase your PR please? Jut fixed the action issue.

specs/Modal.events.spec.js
diasbruno: Ah now I remember it...

The `mouse down and up` are to test the scenario where you click and hold on a content and release (mouse up) on the overlay.

ErikDeviashin:
> @ErikDeviashin Can you please rebase your PR please? Jut fixed the action issue.

Sure, did it

specs/Modal.events.spec.js
ErikDeviashin: Yes that's it!
It just didn't work when I ran the tests, so I compared it to real behavior and suggesting a way to fix the tests. At least now they works for me :)

diasbruno: I guess in the end the result is correct, but I'm not sure if the test "examplifies" the behavior.

I need to think about it.

specs/Modal.events.spec.js
```diff
@@ -222,6 +222,7 @@ export default () => {
       withModal(props, null, modal => {
         mouseDownAt(moverlay(modal));
         mouseUpAt(mcontent(modal));
+        clickAt(moverlay(modal));
```
diasbruno, line 225: Here is the behavoir I was talking about. 

This `clickAt` shouldn't be here.

specs/Modal.events.spec.js
diasbruno: This one too.

If I press down on the content, drag to the overlay, it must not close the modal.

diasbruno:
```suggestion
```

specs/Modal.events.spec.js
diasbruno:
```suggestion
```

specs/Modal.events.spec.js
```diff
@@ -236,9 +237,40 @@ export default () => {
...
+      withModal(
+        {
+          isOpen: true,
+          onRequestClose: requestCloseCallback
+        },
+        <button ref={innerButtonRef} onClick={click} />,
+        modal => {
+          // imitate regular click with all mouse events on button inside modal
+          mouseDownAt(innerButton);
+          mouseUpAt(innerButton);
```
diasbruno, line 264:
```suggestion
```

specs/Modal.events.spec.js
```diff
@@ -236,9 +237,40 @@ export default () => {
...
+          clickAt(innerButton);
+          // imitate regular click with all mouse events on modal overlay
+          mouseDownAt(moverlay(modal));
+          mouseUpAt(moverlay(modal));
```
diasbruno, line 268:
```suggestio
```

diasbruno:
I've made a few suggestions, I think that is what we are expecting from your new test.

diasbruno:
Those 2 other tests, they need to stay the same to guarantee the behavoir when dragging from the `content -> overlay` and `overlay -> content`.

diasbruno:
Ok. I'll have a look when I have time. 
Thanks so much for taking the time to help with this issue.

khmelevskii:
@diasbruno it would be very cool if this fix will be merged to master. I have the same problem and I can't see any workaround to fix it

diasbruno:
@khmelevskii This patch needs a better look to see if it is right...
If you have time, you can take this patch and give it a try...

What I expect:

- When press and hold on the overlay and release on the content, it must not close
- When press and hold on the content and release on the overlay, it must not close
- See the [https://codesandbox.io/s/react-modal-layout-issue-xgfsvk](https://codesandbox.io/s/react-modal-layout-issue-xgfsvk) an try to replicate it

khmelevskii:
Thank you @diasbruno ! it will be pretty difficult to find time for this soon, but thank you for sharing correct expectations.

ErikDeviashin:
@diasbruno @khmelevskii 
I had to change the solution in our project to just using an extra variable like 
`this.preventOverlayMouseDown = false` (ModalPortal.js 285:42)

This is used like
`this.preventOverlayMouseDown = true;` (in handleContentOnMouseDown func on ModalPortal.js 236:44 line)
and
`if (_this.preventOverlayMouseDown) {
       _this.preventOverlayMouseDown = false;
      } else {
        _this.shouldClose = null;
      }`
(In handleOverlayOnMouseDown func on ModalPortal.js 223:7 line)

This is better just because using `event.stopPropagation();` broke our tooltips a bit, so it might conflict with different libs that use the `mouseDown` event.

@khmelevskii If you need some quick solution look at the `patch-package` library because as you can see it can take a while before it gets fixed. I am using `patch-package` in our project to apply the changes written above

khmelevskii:
Thank you @diasbruno ! I will try

see2ever:
> @diasbruno @khmelevskii I had to change the solution in our project to just using an extra variable like `this.preventOverlayMouseDown = false` (ModalPortal.js 285:42)
> 
> This is used like `this.preventOverlayMouseDown = true;` (in handleContentOnMouseDown func on ModalPortal.js 236:44 line) and `if (_this.preventOverlayMouseDown) { _this.preventOverlayMouseDown = false; } else { _this.shouldClose = null; }` (In handleOverlayOnMouseDown func on ModalPortal.js 223:7 line)
> 
> This is better just because using `event.stopPropagation();` broke our tooltips a bit, so it might conflict with different libs that use the `mouseDown` event.
> 
> @khmelevskii If you need some quick solution look at the `patch-package` library because as you can see it can take a while before it gets fixed. I am using `patch-package` in our project to apply the changes written above

 because those line number are not match the source file, it would be more clear if the patch file pasted here.

PR #1040, opened 2024-02-13
[fixed] Prevent unintended register and unregister #808

### LINKED ISSUE #808: Cannot register modal instance that's already open
### Summary:
When a modal is open, double-clicking on the background overlay results in the modal disappearing then quickly reappearing again with the following warning in the console:

```react_devtools_backend.js:6 React-Modal: Cannot register modal instance that's already open 
    at ModalPortal (http://localhost:4500/static/js/0.chunk.js:60569:5)
    at Modal (http://localhost:4500/static/js/0.chunk.js:60212:5)```

### PR DESCRIPTION
**Pull request by honeymaro**
- Fixed #808
- Checking `this.props.isOpen` instead of `this.state.isOpen` in `componentWillUnmount`.
- Moved the call to `this.beforeOpen()` from the beginning of the `open` method to an else block.
- Set `this.closeTimer` without waiting until the end of the setState.
- Call `this.afterClose()` directly after setting the state in `closeWithoutTimeout`.

Tested environment
- `React` ^18
- `React.StrictMode` / without `React.StrictMode`
(I tested it with reference to https://github.com/reactjs/react-modal/issues/808#issuecomment-1445230192)

This PR also fixes the problem regarding conditional rendering.

Acceptance Checklist:
- [x] Tests
- [x] Documentation and examples (if needed)

Fixes #808

### DIFF
--- a/src/components/ModalPortal.js
+++ b/src/components/ModalPortal.js
@@ -133,7 +133,7 @@ export default class ModalPortal extends Component {
   }
 
   componentWillUnmount() {
-    if (this.state.isOpen) {
+    if (this.props.isOpen) {
       this.afterClose();
     }
     clearTimeout(this.closeTimer);
@@ -227,11 +227,11 @@ export default class ModalPortal extends Component {
   };
 
   open = () => {
-    this.beforeOpen();
     if (this.state.afterOpen && this.state.beforeClose) {
       clearTimeout(this.closeTimer);
       this.setState({ beforeClose: false });
     } else {
+      this.beforeOpen();
       if (this.props.shouldFocusAfterRender) {
         focusManager.setupScopedFocus(this.node);
         focusManager.markForFocusLater();
@@ -268,24 +268,21 @@ export default class ModalPortal extends Component {
 
   closeWithTimeout = () => {
     const closesAt = Date.now() + this.props.closeTimeoutMS;
-    this.setState({ beforeClose: true, closesAt }, () => {
-      this.closeTimer = setTimeout(
-        this.closeWithoutTimeout,
-        this.state.closesAt - Date.now()
-      );
-    });
+    this.setState({ beforeClose: true, closesAt });
+    this.closeTimer = setTimeout(
+      this.closeWithoutTimeout,
+      this.props.closeTimeoutMS
+    );
   };
 
   closeWithoutTimeout = () => {
-    this.setState(
-      {
-        beforeClose: false,
-        isOpen: false,
-        afterOpen: false,
-        closesAt: null
-      },
-      this.afterClose
-    );
+    this.setState({
+      beforeClose: false,
+      isOpen: false,
+      afterOpen: false,
+      closesAt: null
+    });
+    this.afterClose();
   };
 
   handleKeyDown = event => {

### CODE REVIEW

**Comment by diasbruno** [comment]
@honeymaro Looks interesting. Can you test locally with react 16.3+, please? I think 16.3+ should be an ok version to support, and, deprecate the rest.

**Comment by diasbruno** [comment]
And react 17, please.

**Comment by diasbruno** [comment]
Grande lavoro, @honeymaro.

We are going to take some time to stress test this one before we merge, so bear with us.

cc @doeg

**Comment by doeg** [comment]
This is exciting, @honeymaro! 🎉 And thank you for the tag, @diasbruno! I'm happy to take a closer look at this diff over the next couple of days. 

For what it's worth, our enormous codebase isn't on React 18 _yet_ but we're hoping to get there in the near future, so this work is very relevant and appreciated. ❤️ In terms of stress testing, I think it's fair to say that we likely have a _lot_ of "interesting" use cases of those modal so it'll be informative to see how it goes. 😹

**Comment by doeg** [comment]
Thanks all for your patience! 🙏 Just a note to say that I'm testing this out in our codebase now and will report back shortly!

**Comment by doeg** [comment]
**tl;dr: This change seems fine on our end. 👍** 

Below are further notes after testing this change. 

**First**, some initial caveats:
- We use ReactModal in _many_ places (easily hundreds) in many "creative" ways, as I mentioned. (We use conditional rendering of React Modal all over the place, for example.)
- Whenever we upgrade ReactModal, we tend to set aside a week or more to ensure a safe rollout. 
- There are several known issues on our end that are preventing us from upgrading to `react-modal@3.16.1` (unrelated to this particular change). 

So all to say that the testing I did today, though I did my best to be rigorous, is still not as involved as when we upgrade "for real". :) 

**Second**, I can't seem to reproduce the issue described in https://github.com/reactjs/react-modal/issues/808 in our codebase.

This is surprising, as we use conditional rendering (as shown in this simple bug repro [here](https://github.com/reactjs/react-modal/issues/808#issuecomment-1112305603)) _all_ over the place, even though conditional rendering is not recommended per [this note](https://github.com/reactjs/react-modal/issues/808#issuecomment-1155423058). :face_with_monocle:

To note what I did:
- I installed react@18.2.0 and react-dom@18.2.0 and react-modal@13.6.1 (the latest published version, without this fix)
- I opened and closed various modals that use conditional rendering.
  - The expected `Cannot register modal instance that's already open` warning did not appear in the console
  - The `ReactModal__Body--open` was correctly added to `body` element on modal open and removed on modal close.

For good measure, I added a component that copies the bug repro as-written (including `Modal.setAppElement('body')`, which we already do) and was still unable to reproduce the warning. 🤔 I'm not yet sure where the discrepancy is. 

**Third,** we're not yet on React 18 and there are several known issues in our codebase that need to be addressed prior to us upgrading. 

That caveat aside, I checked out @honeymaro's branch and did dev builds against React 18, React 17, and React 16. 

I did notice a few issues but, on further inspection, they happen on `react-modal@3.16.1` (i.e., the latest published version **without** this fix) so... that's more fun stuff for us to fix before we upgrade (😹) but not a blocker for this PR.  

**Overall,** this looks good to me! 🎉

File: src/components/ModalPortal.js [hunk 1]
```diff
@@ -133,7 +133,7 @@ export default class ModalPortal extends Component {
   }
 
   componentWillUnmount() {
-    if (this.state.isOpen) {
+    if (this.props.isOpen) {
```
doeg, line 136: Perhaps a question for you, @diasbruno, as it's not strictly related to this change -- it looks like the only actual use of `this.state.isOpen` is this block:

```jsx
    // Focus only needs to be set once when the modal is being opened
    if (
      this.props.shouldFocusAfterRender &&
      this.state.isOpen &&
      !prevState.isOpen
    ) {
      this.focusContent();
    }
```

Is that correct? 🤔 Is there a way we could accomplish the programmatic focus of the content without maintaining a `this.state.isOpen` variable alongside `this.props.isOpen`? The simplification may be nice, if so.

diasbruno: The state inside the modal is used to manage the timed transition (sync'd with CSS).

When the isOpen flag is changed from true to false, the internal state is true. 
react-modal will wait the css transition N milliseconds, then set to false
(same when you are openning).
When the transition has finished, we then do the tear down any other resources
(applied css names to document.body and others).

I always wanted to implement the transition mechanism using css transition events, 
but never had a change to see the support for it.

doeg: Ahh! Obrigado pela explicação. :)

diasbruno: > I always wanted to implement the transition mechanism using css transition events,
but never had a change to see the support for it.

It would be more reliable than setting timers to manage the internal state.

doeg: That would be cool! I did see this v4 issue about that: https://github.com/reactjs/react-modal/issues/993

gillesbouvier-qz: This proposed fix does not work if the modal is closed with a timeout. `isOpen` will be `false` when we get here, so `afterClose()` won't run and the body class will be left on the DOM.

gillesbouvier-qz: Are you foregoing support for this?
```TSX
isOpen && (
  <Modal
    isOpen={isOpen}
    closeTimeoutMS={300}
    onRequestClose={() => {
      setIsOpen(false);
    }}
    ...
  />
)
```

If you still want to support the above, I believe `isOpen` will be `false` during unmounting and `afterClose()` won't run leaving the body class in place. Using the `isOpen` props seems unsafe to me, as only the modal component should control when it is open or closed.

Maybe I'm missing something?

gillesbouvier-qz: Wouldn't this solve both problems at once?
```suggestion
    clearTimeout(this.closeTimer);
    if (this.props.isOpen || this.closeTimer) {
      this.closeWithoutTimeout();
    }
```

gillesbouvier-qz: It solves the problem in my testcase:
```
ModalPortal.js:164 CLOSING MODAL WITH TIMEOUT 300
ModalPortal.js:177 CLOSE WITH TIMEOUT
ModalPortal.js:322 UN-MOUNTING
ModalPortal.js:184 CLOSE WITHOUT TIMEOUT
ModalPortal.js:100 AFTER CLOSE
```

the idea being that we always want `afterClose()` to run at least once even when unmounting.

scvsoft-cruzfernandez:
Any news on this one? I think my application also needs this.

gillesbouvier-qz:
Here is a patch for `react-modal 3.16.1` that seems to work well for both testing and production:
```PATCH
diff --git a/node_modules/react-modal/lib/components/ModalPortal.js b/node_modules/react-modal/lib/components/ModalPortal.js
index 07e27aa..e8c7358 100644
--- a/node_modules/react-modal/lib/components/ModalPortal.js
+++ b/node_modules/react-modal/lib/components/ModalPortal.js
@@ -91,6 +91,14 @@ var ModalPortal = function (_Component) {
     };

     _this.afterClose = function () {
+      // Guard against double-invocation: when the close timer and componentWillUnmount
+      // fire in the same synchronous flush (e.g. jest.advanceTimersByTime), both paths
+      // call closeWithoutTimeout() -> afterClose(). The flag makes afterClose idempotent
+      // so deregister() is only called once.
+      if (_this._afterCloseCalled) {
+        return;
+      }
+      _this._afterCloseCalled = true;
       var _this$props = _this.props,
           appElement = _this$props.appElement,
           ariaHideApp = _this$props.ariaHideApp,
@@ -132,11 +140,11 @@ var ModalPortal = function (_Component) {
     };

     _this.open = function () {
-      _this.beforeOpen();
       if (_this.state.afterOpen && _this.state.beforeClose) {
         clearTimeout(_this.closeTimer);
         _this.setState({ beforeClose: false });
       } else {
+        _this.beforeOpen();
         if (_this.props.shouldFocusAfterRender) {
           focusManager.setupScopedFocus(_this.node);
           focusManager.markForFocusLater();
@@ -171,9 +179,8 @@ var ModalPortal = function (_Component) {

     _this.closeWithTimeout = function () {
       var closesAt = Date.now() + _this.props.closeTimeoutMS;
-      _this.setState({ beforeClose: true, closesAt: closesAt }, function () {
-        _this.closeTimer = setTimeout(_this.closeWithoutTimeout, _this.state.closesAt - Date.now());
-      });
+      _this.setState({ beforeClose: true, closesAt: closesAt });
+      _this.closeTimer = setTimeout(_this.closeWithoutTimeout, _this.props.closeTimeoutMS);
     };

     _this.closeWithoutTimeout = function () {
@@ -182,7 +189,8 @@ var ModalPortal = function (_Component) {
         isOpen: false,
         afterOpen: false,
         closesAt: null
-      }, _this.afterClose);
+      });
+      _this.afterClose();
     };

     _this.handleKeyDown = function (event) {
@@ -273,6 +281,7 @@ var ModalPortal = function (_Component) {
       beforeClose: false
     };

+    _this._afterCloseCalled = false;
     _this.shouldClose = null;
     _this.moveFromContentToOverlay = null;
     return _this;
@@ -313,15 +322,16 @@ var ModalPortal = function (_Component) {
   }, {
     key: "componentWillUnmount",
     value: function componentWillUnmount() {
-      if (this.state.isOpen) {
-        this.afterClose();
-      }
       clearTimeout(this.closeTimer);
+      if (this.props.isOpen || this.state.isOpen) {
+        this.closeWithoutTimeout();
+      }
       cancelAnimationFrame(this.openAnimationFrame);
     }
   }, {
     key: "beforeOpen",
     value: function beforeOpen() {
+      this._afterCloseCalled = false;
       var _props = this.props,
           appElement = _props.appElement,
           ariaHideApp = _props.ariaHideApp,
```

PR #1043, opened 2024-02-27
WIP: rework build system to use esbuild.

### LINKED ISSUE #1036: Moving to a better build system (hardcore level)
react-modal uses an old toolchain system for build and testing (not that the tools aren't great, 
but because javascript toolchain are pretty messy and not reliable as things changes all the time).

The idea is to move to a "modern" build system (probably vite as it also has compatibility with karma, 
but suggestions are welcome) as people apparently have move to it.

diasbruno:
Fixes #1036.

This is an update to the build system to use esbuild.

It need to:

- [x] Build a pre-compiled version
- [ ] Build a minified version
- [x] Build an esm module
- [ ] Build an cjs module
- [ ] Run dev/examples server
- [ ] Run tests


shell.nix
```diff
@@ -0,0 +1,6 @@
+{ nixpkgs ? import (fetchTarball("channel:nixos-23.11")) {} }:
+let
+  inherit (nixpkgs) pkgs;
+in pkgs.mkShell {
+  nativeBuildInputs = with pkgs; [jq nodejs_20 swc nodePackages.typescript-language-server];
```
diasbruno, line 5:
```suggestion
  nativeBuildInputs = with pkgs; [jq nodejs_20 nodePackages.typescript-language-server];
```

diasbruno:
@doeg What do you think about removing "Build an esm/cjs modules"? 

I'm thinking in reducing the amount of artifacts and just a provide a minified version of the library, and, point to main file to "src/index.[tj]s" (this might fix the issue of using  fork of this project.

diasbruno:
We used to pre-compile the library because of some webpack loaders and babel transformers, but I think that if we move to functions and hooks, we can remove build step.

doeg:
@diasbruno to tell you the truth, I haven't maintained a React library before and can't give you an informed answer from that perspective. 🤔 

Please correct me if I'm wrong: my understanding is that CommonJS modules are typically intended for backend/node imports. I looked at Next.js as an example of server-side imports and it looks like even they support ESM these days. I agree that CommonJS is probably not the best fit for a front-end library like this one. 

A question, though! 🙇 If the module distributes just the minified version of the library, wouldn't the entrypoint file still need to use ESM module syntax (or equivalent)?

package.json
```diff
@@ -23,46 +23,18 @@
...
-    "karma-sourcemap-loader": "^0.3.8",
-    "karma-webpack": "^2.0.4",
-    "mocha": "^8.4.0",
-    "npm-run-all": "^4.1.1",
-    "prettier": "^1.19.1",
+    "esbuild": "0.20.0",
+    "esbuild-dev-server": "^0.3.0",
+    "esbuild-node-externals": "^1.13.0",
+    "eslint": "^8.56.0",
+    "eslint-plugin-react": "^7.33.2",
```
doeg, line 33: This diff is amazing! 😂

diasbruno:
>  If the module distributes just the minified version of the library, wouldn't the entrypoint file still need to use ESM module syntax (or equivalent)?

Depends on which version of the language (ES5, ES2015, ...) we compile it to.

If we don't pre-compile the library, users needs to have all the correct languages features to compile (classes, arrows and so on). But, if we move away from class components - I'm guessing, we won't have problems.

I'm thinking on release this as `v4-rc1`, so we can give it a try and see how that works.

diasbruno:
I'm going to drop `esm/cjs` for the first release candidate, so we can test it early.

doeg:
Makes sense to me! Let me know if I can be helpful in testing or otherwise. 😸

rdtabb:
@diasbruno  hi! I wonder how the migration to esbuild is going? Can I help?

I really like your lib and have been using it extensively for more than a year in prod at my job, then I saw this pr, and thought maybe I can be of some use. I have a pretty extensive experience in migrating legacy babel/webpack/plain-ol'-tsc build systems to swc/esbuild/vite-lib-mode etc. I also am quite experienced with maintaining react-js libraries (although all of them are private repos in my company).

So if you can bring me up to speed and maybe offload some work to me, I would be more than happy!

(i don't want to be a nuisance and just try to contribute stuff, who knows, maybe you already got everything figured out, he-he)

diasbruno:
@rdtabb What made me pause the whole update process is that karma has been deprecated. We can't run the tests using `jsdom`,  since we need to tweak some browser behavior (specially regarding "tabbing" inside the modal). 

I've tried a few alternatives, but the process was really stressful.

If you want a really hard *quest*, you can take this branch (or create another one to give it a try). The goal would be to build and test on a real browser.

If you want to give it a try, I can help/assist with anything you need. Just let me know.

PR #1045, opened 2024-03-08
Fixing issue with tabbing when a radio is the last element of the modal

### LINKED ISSUE #636: Focus escapes modal when radio buttons in modal
### Summary:

When there are radio buttons in a modal, you can escape focus with shift-tab.

### Steps to reproduce:

Steps are included in the test case below

### Expected behavior:

You should not be able to escape the modal's focus

### Link to example of issue:

https://codesandbox.io/s/my9vm34m68

### Additional notes:

Not sure if it's related specifically to radio buttons, but that is what is causing the issue for me

loganscharen:
The issue here was that the code was checking if the activeElement was the last element, but for radio groups, you can tab out without it being the last element. This checks if the tail and activeElement are part of the same radio group and if so treats it as if it were the last element

Acceptance Checklist:
- [X] Tests
- [ ] Documentation and examples (if needed)

Fixes #636 .

### DIFF
--- a/src/helpers/scopeTab.js
+++ b/src/helpers/scopeTab.js
@@ -7,7 +7,7 @@ function getActiveElement(el = document) {
 }
 
 function areFromSameRadioGroup(a, b) {
-  return a.type === "radio" && b.type === "radio" && a.name === b.name;
+  return a.name === b.name && a.type === "radio" && b.type === "radio";
 }
 
 export default function scopeTab(node, event) {


src/helpers/scopeTab.js
```diff
@@ -29,7 +29,13 @@ export default function scopeTab(node, event) {
     target = tail;
   }
 
-  if (tail === activeElement && !shiftKey) {
+  if (
+    (tail === activeElement ||
+       (activeElement.type === "radio" &&
+         tail.type === "radio" &&
+         tail.name === activeElement.name)) &&
```
diasbruno, line 36: You can move this into a function.

loganscharen:
Moved that to a function and updated the basic form example to have the radios at the end to be able to see the functionality

src/helpers/scopeTab.js
```diff
@@ -6,6 +6,14 @@ function getActiveElement(el = document) {
     : el.activeElement;
 }
 
+function isActiveElementInTailRadioGroup(activeElement, tail) {
```
diasbruno, line 9:
```suggestion
function areFromSameRadioGroup(a, b) {
```

src/helpers/scopeTab.js
```diff
@@ -29,7 +37,11 @@ export default function scopeTab(node, event) {
     target = tail;
   }
 
-  if (tail === activeElement && !shiftKey) {
+  if (
+    (tail === activeElement ||
+       isActiveElementInTailRadioGroup(activeElement, tail)) &&
```
diasbruno, line 42:
```suggestion
       areFromSameRadioGroup(activeElement, tail)) &&
```

examples/basic/forms/index.js
```diff
@@ -45,21 +45,21 @@ class Forms extends Component {
                 <input type="text"  />
```
diasbruno, line 45: You can revert this file. The tests is what matters.

diasbruno:
@loganscharen @doeg Is there any accessibility tool or documentation where we chan check this behavior?

diasbruno:
> Do you mean checking whether the new behaviour is compliant with W3 standards?

Yep, please.

loganscharen:
Is this what you mean?
https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/#keyboardinteraction

diasbruno:
@loganscharen Yas, but for the case of radio groups. If we are going to intervening on default browser behavior, we need to do it right.

loganscharen:
https://www.w3.org/WAI/ARIA/apg/patterns/radio/#keyboardinteraction

So this? That tab is used to go into and out of the radio group and not between the elements of the group. So the radio group is the last tabbable element, not each radio button separately.

loganscharen:
@diasbruno anything else needed to get this merged?

diasbruno:
@loganscharen Thanks. That's what I was looking for.

There are some rules on that specification. Should we add tests to check if the behavior is correct (arrows, space)?

Also, long time ago there was this issue https://github.com/reactjs/react-modal/blob/master/src/helpers/scopeTab.js#L48, but I think I've implemented not respecting the radio specification. Should this cause this bug?

loganscharen:
I don't see why we would need to test that space and arrow are working correctly, we're not touching that functionality at all.

The safari bug doesn't have any impact on this, this is just occurring because we're currently looking at the last element, but from a radio group, the last element isn't always the last tabbable element.

diasbruno:
Ok. I'll have a look on it later.

src/helpers/scopeTab.js
```diff
@@ -7,11 +7,7 @@ function getActiveElement(el = document) {
 }
 
 function areFromSameRadioGroup(a, b) {
-  return (
-    activeElement.type === "radio" &&
-    tail.type === "radio" &&
-    tail.name === activeElement.name
-  );
+  return a.type === "radio" && b.type === "radio" && a.name === b.name;
```
diasbruno, line 10:
```suggestion
  return a.name === b.name && a.type === "radio" && b.type === "radio";
```

loganscharen:
@diasbruno Forgot about this, but any chance you could take a look at it again?

Divya1k5:
@loganscharen @diasbruno This issue still exists. Any reasons why y'all have held off on merging this PR?

Issue #1056, opened 2025-01-29
dragging  property of ant draggable modal not woking when left is applied


sneha-soxo:
In antd draggable modal when i applied left:0 as inline style for modal, the dragging property not working.

fluxaro:
i would like to fix this

diasbruno:
Need more information about this.

PR #1077, opened 2026-03-05
fix: allow draggable modal to work with inline left style #1056

fluxaro:
### What this PR does
This fixes the issue where draggable modals stop working when an inline `left: 0` style is applied (Issue #1056).

### How it was fixed
- Updated the draggable logic to correctly calculate offsets even when `left` is set inline.
- Ensures the modal can still be dragged as expected.
- Preserves all existing modal functionality.

### Testing
- Verified dragging works with `left: 0`.
- Verified dragging still works with other `left` values.
- Checked that no other modal styles are broken.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "example": "examples"
   },
   "scripts": {
-    "start": "npx webpack-dev-server --config ./scripts/webpack.config.js --inline --host 127.0.0.1 --content-base examples/",
+    "start": "set NODE_OPTIONS=--openssl-legacy-provider && npx webpack-dev-server --config ./scripts/webpack.config.js --inline --host 127.0.0.1 --content-base examples/",
     "test": "cross-env NODE_ENV=test karma start",
     "lint": "eslint src/"
   },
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "example": "examples"
   },
   "scripts": {
-    "start": "npx webpack-dev-server --config ./scripts/webpack.config.js --inline --host 127.0.0.1 --content-base examples/",
+    "start": "set NODE_OPTIONS=--openssl-legacy-provider && npx webpack-dev-server --config ./scripts/webpack.config.js --inline --host 127.0.0.1 --content-base examples/",
```
diasbruno, line 17: Please remove.

src/index.js
```diff
@@ -1,3 +1,5 @@
 import Modal from "./components/Modal";
+import DraggableModal from "./components/DraggableModal";
 
 export default Modal;
+export { DraggableModal };
```
diasbruno, line 5: `DraggableModal` shouldn't be exported by this file.

